PR #767, opened 2025-07-23

AbdeleahmanEl-Desoky:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change


modules/Attendance/Presenters/AttendancePresenter.php
```diff
@@ -106,6 +106,7 @@ public function present(bool $isListing = false): array
                 'attendance_constraint'=> $this->attendance->user->professionalData->attendanceConstraint ?[
                     'id'=> (string) $this->attendance->user->professionalData->id,
                     'constraint_name'=> $this->attendance->user->professionalData->attendanceConstraint?->constraint_name,
+                    'constraint_type'=> $this->attendance->user->professionalData->attendanceConstraint?->constraint_type,
                     'constraint_config'=> $this->attendance->user->professionalData?->attendanceConstraint->constraint_config,
```
mabou7agar, line 110: Potential null reference issue: `constraint_config` is accessed directly without null check (`?->`) like the other properties above it. Should be consistent with the null-safe operator.

PR #768, opened 2025-07-23

AbdeleahmanEl-Desoky:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/modules/Attendance/Filters/AttendanceFilter.php
+++ b/modules/Attendance/Filters/AttendanceFilter.php
@@ -175,8 +175,8 @@ class AttendanceFilter extends SearchModelFilter
 
     public function constraint($constraintId)
     {
-       return $this->whereHas('appliedConstraints', function ($query) use ($constraintId) {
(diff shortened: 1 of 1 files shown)


modules/Attendance/Filters/AttendanceFilter.php
```diff
@@ -175,8 +175,8 @@ public function branch($branchId)
 
     public function constraint($constraintId)
     {
-       return $this->whereHas('appliedConstraints', function ($query) use ($constraintId) {
-            $query->where('constraint_id', $constraintId);
+       return $this->whereHas('professionalData', function ($query) use ($constraintId) {
+            $query->where('attendance_constraint_id',$constraintId);
```
mabou7agar, line 179: The column name `attendance_constraint_id` should be consistent with the naming convention used in the rest of the codebase. Please verify if this is the correct column name or if it should follow a different pattern.

PR #771, opened 2025-07-23

AbdeleahmanEl-Desoky:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/modules/Attendance/Presenters/AttendancePresenter.php
+++ b/modules/Attendance/Presenters/AttendancePresenter.php
@@ -71,7 +71,7 @@ class AttendancePresenter extends AbstractPresenter
                 'email' => $this->attendance->user->email,
                 'birthdate' => $this->attendance->user?->companyUser?->birthdate_gregorian ?? null,
                 'country' => $this->attendance->user->companyUser?->country?->name ?: null,
-                'gender' => $this->attendance->user->companyUser->gender,
+                'gender' => __('validation.' . $this->attendance->user->companyUser->gender),
                 'phone' => $this->attendance->user->companyUser->phone,
             ] : null,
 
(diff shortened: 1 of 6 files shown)


modules/Attendance/Presenters/AttendancePresenter.php
```diff
@@ -71,7 +71,7 @@ public function present(bool $isListing = false): array
                 'email' => $this->attendance->user->email,
                 'birthdate' => $this->attendance->user?->companyUser?->birthdate_gregorian ?? null,
                 'country' => $this->attendance->user->companyUser?->country?->name ?: null,
-                'gender' => $this->attendance->user->companyUser->gender,
+                'gender' => __('validation.' . $this->attendance->user->companyUser->gender),
```
mabou7agar, line 74: The translation key `validation.` prefix might not be appropriate here since gender is not a validation attribute. Consider using a more suitable translation key prefix like `genders.` or `profile.`.

modules/User/Models/User.php
```diff
@@ -33,6 +33,14 @@
 
 //use BasePackage\Shared\Traits\HasTranslations;
 
+/**
+ * @property string $id
```
mabou7agar, line 37: The PHPDoc block is incomplete. It should include all properties from the model including those inherited from parent classes and traits. Also, consider using `@method` tags for relations since they are dynamically added by Laravel.

modules/User/Models/User.php
```diff
@@ -33,6 +33,14 @@
...
+ * @property string $name
+ * @property string $email
+ * @property string|null $phone
+ * @property-read CompanyUser|null $companyUser
+ * @property-read UserProfessionalData|null $professionalData
```
mabou7agar, line 42: Missing `@property-read` for other relations that exist in the User model (e.g., roles, permissions, etc.). The docblock should be comprehensive.

PR #776, opened 2025-07-24
fix AuthService

AbdeleahmanEl-Desoky:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/modules/Auth/Services/AuthService.php
+++ b/modules/Auth/Services/AuthService.php
@@ -4,8 +4,9 @@ namespace Modules\Auth\Services;
 
 use App\Exceptions\CustomException;
 use Carbon\Carbon;
-use Faker\Core\Uuid;
+
 use Ichtrojan\Otp\Otp;
+use Ramsey\Uuid\Uuid;
 use Illuminate\Support\Facades\Hash;
 use Modules\Auth\Commands\ResendOtpCommand;
 use Modules\Auth\Commands\ResetPasswordCommand;
@@ -55,7 +56,7 @@ class AuthService
         $isContinueWithOTP = $this->settingCRUDService->getValue('continue_with_otp');
         if ($isContinueWithOTP) {
             $user = $this->userRepository->getUserByEmail($authDTO->getEmail());
-            $this->sendOtpEmail->loginWithOtp($user->id);
+            $this->sendOtpEmail->loginWithOtp(Uuid::fromString($user->id));
             return [null, $user];
         }
 
@@ -63,7 +64,7 @@ class AuthService
         if (!$token) {
             throw new \ErrorException(__("validation.invalid-credential"), 403);
         }
-        $user = auth()->user();
+        $user = JWTAuth::user();
         return [$token, $user];
     }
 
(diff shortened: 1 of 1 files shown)


modules/Auth/Services/AuthService.php
```diff
@@ -4,8 +4,9 @@
 
 use App\Exceptions\CustomException;
 use Carbon\Carbon;
-use Faker\Core\Uuid;
+
```
mabou7agar, line 7: The import of `Faker\Core\Uuid` was removed but not replaced with `Ramsey\Uuid\Uuid` in the same line. This could lead to confusion in the diff. Consider keeping the removal and addition as separate changes for better clarity.

modules/Auth/Services/AuthService.php
```diff
@@ -99,7 +100,7 @@
         if ((new Otp)->validate($validateOtpDTO->getIdentifier(), $validateOtpDTO->getOtp())->status == true) {
             $user = $this->userCRUDService->getUserByIdentifier($validateOtpDTO->getIdentifier());
 
-            $token = $this->verficationDataRepository->createToken($user->id, ["can_reset_password" => true])->token;
+            $token = $this->verficationDataRepository->createToken(Uuid::fromString($user->id), ["can_reset_password" => true])->token;
```
mabou7agar, line 103: The method `Uuid::fromString()` is used but there's no check if `$user->id` is actually a valid UUID string. This could lead to runtime errors if the ID is not in UUID format. Consider adding validation or handling potential exceptions.

modules/Auth/Services/AuthService.php
```diff
@@ -161,12 +162,14 @@
 
     private function checkOtpByStep($step, $identifier, $otp)
     {
```
mabou7agar, line 164: The condition change from `empty($step)` to `!property_exists($step, 'login_option')` is not equivalent. The original check was for empty step, while the new one checks for a property. This could lead to different behavior. Consider if this is the intended logic change.

modules/Auth/Services/AuthService.php
```diff
@@ -161,12 +162,14 @@
...
-        if (empty($step)) {
+        if (!property_exists($step, 'login_option')) {
             return true;
         }
+
```
mabou7agar, line 168: Unnecessary empty line added. This doesn't improve readability.

modules/Auth/Services/AuthService.php
```diff
@@ -161,12 +162,14 @@
...
         if ($step->login_option == "otp") {
             return (new Otp)->validate($identifier, $otp)->status;
         }
+
```
mabou7agar, line 172: Unnecessary empty line added. This doesn't improve readability.

modules/Auth/Services/AuthService.php
```diff
@@ -211,7 +214,7 @@ public function getLoginWays(GetLoginWaysDTO $getLoginWaysDTO)
         }
 
         $this->sendOtpByStep($step, $getLoginWaysDTO->getIdentifier());
-        $token = $this->verficationDataRepository->createToken($user->id, ["order" => 1, "login_way" => $loginWay])->token;
+        $token = $this->verficationDataRepository->createToken(Uuid::fromString($user->id), ["order" => 1, "login_way" => $loginWay])->token;
```
mabou7agar, line 217: The `Uuid::fromString()` conversion should be validated first to ensure `$user->id` is actually a valid UUID string to prevent potential runtime errors.

modules/Auth/Services/AuthService.php
```diff
@@ -316,7 +319,7 @@ public function checkQuestionAnswer(QuestionVerificationDTO $questionVerificatio
                 return [false, null];
             }
         }
-        $verficationData = $this->verficationDataRepository->createToken($user->id, ["change_email" => 1]);
+        $verficationData = $this->verficationDataRepository->createToken(Uuid::fromString($user->id), ["change_email" => 1]);
```
mabou7agar, line 322: There's a typo in the variable name `verficationData` (missing an 'r' in 'verification'). Also, the same typo exists in the method name `verficationDataRepository` which should be `verificationDataRepository`.

PR #778, opened 2025-07-24
Feature/attendance system

AbdeleahmanEl-Desoky:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/modules/Auth/Services/AuthService.php
+++ b/modules/Auth/Services/AuthService.php
@@ -4,9 +4,8 @@ namespace Modules\Auth\Services;
 
 use App\Exceptions\CustomException;
 use Carbon\Carbon;
-
+use Faker\Core\Uuid;
 use Ichtrojan\Otp\Otp;
-use Ramsey\Uuid\Uuid;
 use Illuminate\Support\Facades\Hash;
 use Modules\Auth\Commands\ResendOtpCommand;
 use Modules\Auth\Commands\ResetPasswordCommand;
@@ -56,7 +55,7 @@ class AuthService
         $isContinueWithOTP = $this->settingCRUDService->getValue('continue_with_otp');
         if ($isContinueWithOTP) {
             $user = $this->userRepository->getUserByEmail($authDTO->getEmail());
-            $this->sendOtpEmail->loginWithOtp(Uuid::fromString($user->id));
+            $this->sendOtpEmail->loginWithOtp($user->id);
             return [null, $user];
         }
 
@@ -64,7 +63,7 @@ class AuthService
         if (!$token) {
             throw new \ErrorException(__("validation.invalid-credential"), 403);
         }
-        $user = JWTAuth::user();
+        $user = auth()->user();
         return [$token, $user];
     }
 
@@ -100,7 +99,7 @@ class AuthService
         if ((new Otp)->validate($validateOtpDTO->getIdentifier(), $validateOtpDTO->getOtp())->status == true) {
             $user = $this->userCRUDService->getUserByIdentifier($validateOtpDTO->getIdentifier());
 
-            $token = $this->verficationDataRepository->createToken(Uuid::fromString($user->id), ["can_reset_password" => true])->token;
+            $token = $this->verficationDataRepository->createToken($user->id, ["can_reset_password" => true])->token;
             return $token;
         }
         throw new \ErrorException(__("validation.invalid-otp"), 401);
@@ -110,10 +109,10 @@ class AuthService
     public function ResetPassword(ResetPasswordCommand $resetPasswordCommand)
     {
         $token = $this->verficationDataRepository->findOneBy(["token" => $resetPasswordCommand->getToken()]);
-        if ($token && isset($token->data["can_reset_password"]) && $token->data["can_reset_password"] == 1) {
+        if (true || $token && isset($token->data["can_reset_password"]) && $token->data["can_reset_password"] == 1) {
             $user = $this->userCRUDService->getUserByIdentifier($resetPasswordCommand->getIdentifier());
 
-            $this->userRepository->updateUser(Uuid::fromString($user->id), ["password" => $resetPasswordCommand->getPassword()]);
+            $this->userRepository->updateUser($user->id, ["password" => $resetPasswordCommand->getPassword()]);
 
             return $this;
         }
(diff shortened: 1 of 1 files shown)


modules/Auth/Services/AuthService.php
```diff
@@ -4,9 +4,8 @@
 
 use App\Exceptions\CustomException;
 use Carbon\Carbon;
-
+use Faker\Core\Uuid;
```
mabou7agar, line 7: The `Faker\Core\Uuid` namespace is being used to replace `Ramsey\Uuid\Uuid`. Please ensure this is intentional as Faker's UUID generator is typically used for testing/mocking purposes, while Ramsey/Uuid is the standard for production UUID generation.

modules/Auth/Services/AuthService.php
```diff
@@ -56,15 +55,15 @@
         $isContinueWithOTP = $this->settingCRUDService->getValue('continue_with_otp');
         if ($isContinueWithOTP) {
             $user = $this->userRepository->getUserByEmail($authDTO->getEmail());
-            $this->sendOtpEmail->loginWithOtp(Uuid::fromString($user->id));
+            $this->sendOtpEmail->loginWithOtp($user->id);
```
mabou7agar, line 58: The change removes `Uuid::fromString()` conversion but doesn't ensure `$user->id` is already in UUID format. This could lead to inconsistencies if the ID format varies.

modules/Auth/Services/AuthService.php
```diff
@@ -56,15 +55,15 @@
...
             return [null, $user];
         }
 
         $token = JWTAuth::attempt($authDTO->toArray());
         if (!$token) {
             throw new \ErrorException(__("validation.invalid-credential"), 403);
         }
-        $user = JWTAuth::user();
+        $user = auth()->user();
```
mabou7agar, line 66: Using `auth()->user()` instead of `JWTAuth::user()` is fine, but ensure consistency across the codebase - don't mix both approaches.

modules/Auth/Services/AuthService.php
```diff
@@ -110,10 +109,10 @@
     public function ResetPassword(ResetPasswordCommand $resetPasswordCommand)
     {
         $token = $this->verficationDataRepository->findOneBy(["token" => $resetPasswordCommand->getToken()]);
-        if ($token && isset($token->data["can_reset_password"]) && $token->data["can_reset_password"] == 1) {
+        if (true || $token && isset($token->data["can_reset_password"]) && $token->data["can_reset_password"] == 1) {
```
mabou7agar, line 112: The condition `if (true || ...)` appears to be a debugging leftover. This will bypass all security checks and should be removed before merging.

modules/Auth/Services/AuthService.php
```diff
@@ -110,10 +109,10 @@
...
             $user = $this->userCRUDService->getUserByIdentifier($resetPasswordCommand->getIdentifier());
 
-            $this->userRepository->updateUser(Uuid::fromString($user->id), ["password" => $resetPasswordCommand->getPassword()]);
+            $this->userRepository->updateUser($user->id, ["password" => $resetPasswordCommand->getPassword()]);
```
mabou7agar, line 115: Changing from `Uuid::fromString($user->id)` to just `$user->id` might cause issues if the ID isn't already in UUID format. Ensure this change is intentional and properly tested.

modules/Auth/Services/AuthService.php
```diff
@@ -162,28 +161,28 @@ private function sendOtpByStep($step, $identifier)
 
     private function checkOtpByStep($step, $identifier, $otp)
     {
-        if (!property_exists($step, 'login_option')) {
+        if (empty($step)) {
```
mabou7agar, line 164: The condition `if (empty($step))` is too broad and could lead to unintended behavior. Consider checking specifically for null or invalid step objects like done in line 178.

modules/Auth/Services/AuthService.php
```diff
@@ -162,28 +161,28 @@ private function sendOtpByStep($step, $identifier)
...
-    //         throw new \Exception(__("validation.invalid-otp"), 401);
-    //     }
-    //     return true;
-    // }
+
     private function checkPasswordByStep($step, $identifier, $password)
     {
         $user = $this->userCRUDService->getUserByIdentifier($identifier);
 
+        if ($step === null || !is_object($step) || get_class($step) === 'stdClass') {
```
mabou7agar, line 178: The check `get_class($step) === 'stdClass'` is too specific and might not cover all cases. Consider using `instanceof` with a proper class name if available, or remove this check if not necessary.

modules/Auth/Services/AuthService.php
```diff
@@ -162,28 +161,28 @@ private function sendOtpByStep($step, $identifier)
...
+            return false;
+        }
+
+        if (!property_exists($step, 'login_option')) {
```
mabou7agar, line 182: The `property_exists` check should be moved before the password verification logic to fail fast if the step is invalid.

modules/Auth/Services/AuthService.php
```diff
@@ -214,7 +213,7 @@ public function getLoginWays(GetLoginWaysDTO $getLoginWaysDTO)
         }
 
         $this->sendOtpByStep($step, $getLoginWaysDTO->getIdentifier());
-        $token = $this->verficationDataRepository->createToken(Uuid::fromString($user->id), ["order" => 1, "login_way" => $loginWay])->token;
+        $token = $this->verficationDataRepository->createToken($user->id, ["order" => 1, "login_way" => $loginWay])->token;
```
mabou7agar, line 216: The change removes the `Uuid::fromString()` conversion but doesn't ensure `$user->id` is already in the correct format. This could lead to inconsistencies if `$user->id` is not properly formatted as a UUID string elsewhere.

modules/Auth/Services/AuthService.php
```diff
@@ -290,7 +289,7 @@ public function loginBySteps(LoginStepDTO $loginStepDTO)
         //get next step
 
         if ($nextStep) {//if we have step
-            $token = $this->verficationDataRepository->createToken(Uuid::fromString($user->id), ["order" => $verficationData->data["order"] + 1, "login_way" => $loginWay])->token;
+            $token = $this->verficationDataRepository->createToken($user->id, ["order" => $verficationData->data["order"] + 1, "login_way" => $loginWay])->token;
```
mabou7agar, line 292: The change from `Uuid::fromString($user->id)` to `$user->id` might cause issues if `$user->id` is not already a UUID string. Ensure the ID is properly formatted as a UUID string before passing it.

modules/Auth/Services/AuthService.php
```diff
@@ -319,7 +318,7 @@ public function checkQuestionAnswer(QuestionVerificationDTO $questionVerificatio
                 return [false, null];
             }
         }
-        $verficationData = $this->verficationDataRepository->createToken(Uuid::fromString($user->id), ["change_email" => 1]);
+        $verficationData = $this->verficationDataRepository->createToken($user->id, ["change_email" => 1]);
```
mabou7agar, line 321: There's a typo in the variable name `verficationData` - it should be `verificationData` (missing an 'i').

mabou7agar, line 321: The change from `Uuid::fromString($user->id)` to just `$user->id` might cause issues if the system expects a UUID object. Ensure this change is intentional and properly tested.

PR #781, opened 2025-07-24
rollback all commits

AbdeleahmanEl-Desoky:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/modules/Auth/Services/AuthService.php
+++ b/modules/Auth/Services/AuthService.php
@@ -161,11 +161,8 @@ class AuthService
 
     private function checkOtpByStep($step, $identifier, $otp)
     {
-        if (empty($step)) {
-            return true;
-        }
-        if ($step->login_option == "otp") {
-            return (new Otp)->validate($identifier, $otp)->status;
+        if ($step->login_option == "otp" && (new Otp)->validate($identifier, $otp)->status == false) {
+            throw new \Exception(__("validation.invalid-otp"), 401);
         }
         return true;
     }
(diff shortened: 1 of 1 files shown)


modules/Auth/Services/AuthService.php
```diff
@@ -161,11 +161,8 @@ private function sendOtpByStep($step, $identifier)
 
     private function checkOtpByStep($step, $identifier, $otp)
     {
-        if (empty($step)) {
-            return true;
-        }
-        if ($step->login_option == "otp") {
-            return (new Otp)->validate($identifier, $otp)->status;
+        if ($step->login_option == "otp" && (new Otp)->validate($identifier, $otp)->status == false) {
```
mabou7agar, line 164: The logic change here is incorrect. Previously, if `$step` was empty, it would return true (allowing the flow to continue). Now it throws an exception if OTP validation fails, but doesn't handle the empty `$step` case at all. This could lead to unexpected behavior when `$step` is null.

modules/Auth/Services/AuthService.php
```diff
@@ -161,11 +161,8 @@
...
+            throw new \Exception(__("validation.invalid-otp"), 401);
         }
         return true;
```
mabou7agar, line 167: The `return true` statement should be moved inside an else clause or after the OTP validation check to make the control flow clearer. Currently it's ambiguous whether this is the default return or part of the OTP validation logic.

PR #785, opened 2025-07-24
delete duplicate

AmrSaleh2234:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/lang/ar/names.php
+++ b/lang/ar/names.php
@@ -51,6 +51,9 @@ return [
     'organization-list*management' => 'إدارة',
     'job-title' => 'المسمى الوظيفي',
     'job-type' => 'نوع الوظيفة',
+
+    'organization-list*job-title' => 'المسمى الوظيفي',
+    'organization-list*job-type' => 'نوع الوظيفة',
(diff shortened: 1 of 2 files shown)


lang/ar/names.php
```diff
@@ -51,6 +51,9 @@
     'organization-list*management' => 'إدارة',
     'job-title' => 'المسمى الوظيفي',
     'job-type' => 'نوع الوظيفة',
+
+    'organization-list*job-title' => 'المسمى الوظيفي',
```
mabou7agar, line 55: The key `organization-list*job-title` is redundant as it's already defined on line 52 with the same value `المسمى الوظيفي`.

lang/ar/names.php
```diff
@@ -51,6 +51,9 @@
...
+    'organization-list*job-type' => 'نوع الوظيفة',
```
mabou7agar, line 56: The key `organization-list*job-type` is redundant as it's already defined on line 53 with the same value `نوع الوظيفة`.

PR #787, opened 2025-07-24
subEntitypermission

AmrSaleh2234:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/modules/SubEntity/Models/SubEntity.php
+++ b/modules/SubEntity/Models/SubEntity.php
@@ -69,9 +69,9 @@ class SubEntity extends Model
             }
         });
 
-//        static::created(function (self $subEntity) {
-//            $subEntity->createDefaultPermissions();
-//        });
+        static::created(function (self $subEntity) {
+            $subEntity->createDefaultPermissions();
+        });
     }
 
     protected static function generateUniqueSlug(string $name, $ignoreId = null): string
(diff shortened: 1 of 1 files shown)


modules/SubEntity/Models/SubEntity.php
```diff
@@ -175,15 +175,15 @@ protected function createDefaultPermissions(): void
         }
 
         $module = $this->mainProgram->slug;
-        $resource = $this->id;
-        $companyId = tenant("id");
+        $resource = $this->name . '*' . $this->id;
```
mabou7agar, line 178: The change from using `id` to `name * id` for the resource seems arbitrary. Please clarify the reason for this change as it might affect existing permission checks.

modules/SubEntity/Models/SubEntity.php
```diff
@@ -175,15 +175,15 @@ protected function createDefaultPermissions(): void
...
 
         foreach (self::PERMISSION_ACTIONS as $action) {
             Permission::firstOrCreate([
                 'name' => "{$module}.{$resource}.{$action}",
+                "key" => "{$module}.{$resource}.{$action}",
+
             ], [
                 'status' => true,
```
mabou7agar, line 186: The removal of `company_id` assignment should be justified as this could lead to permissions being created without proper tenant isolation.

PR #796, opened 2025-07-25
return is_absent is_holiday

AbdeleahmanEl-Desoky:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/modules/Attendance/Models/Attendance.php
+++ b/modules/Attendance/Models/Attendance.php
@@ -178,6 +178,21 @@ class Attendance extends Model implements Auditable
         }
     }
 
+    public function getClockInTimeAttribute($value)
+    {
+        if ($this->timezone && $value) {
+            return Carbon::parse($value, timezone: 'UTC')->setTimezone($this->timezone);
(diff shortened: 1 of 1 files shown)


modules/Attendance/Models/Attendance.php
```diff
@@ -178,6 +178,21 @@ public function validateStatusTransition(string $newStatus): void
...
+        if ($this->timezone && $value) {
+            return Carbon::parse($value, timezone: 'UTC')->setTimezone($this->timezone);
+        }
+        return $value;
+    }
+
+    public function getClockOutTimeAttribute($value)
+    {
+        if ($this->timezone && $value) {
+            return Carbon::parse($value, 'UTC')->setTimezone($this->timezone);
```
mabou7agar, line 192: The timezone parameter should be consistent with line 184. Either use named parameter `timezone: 'UTC'` or positional parameter in both methods for consistency.

PR #798, opened 2025-07-26
return is_absent is_holiday

AbdeleahmanEl-Desoky:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/modules/Attendance/Filters/AttendanceFilter.php
+++ b/modules/Attendance/Filters/AttendanceFilter.php
@@ -25,12 +25,12 @@ class AttendanceFilter extends SearchModelFilter
         return $this->where('status', $status);
     }
 
-    public function start_date($date)
+    public function startDate($date)
     {
         return $this->whereDate('clock_in_time', '>=', $date);
     }
 
-    public function end_date($date)
+    public function endDate($date)
     {
         return $this->whereDate('clock_in_time', '<=', $date);
     }
--- a/modules/Attendance/Presenters/AttendanceTeamPresenter.php
+++ b/modules/Attendance/Presenters/AttendanceTeamPresenter.php
@@ -18,6 +18,11 @@ class AttendanceTeamPresenter extends AbstractPresenter
 
     public function present(bool $isListing = false): array
     {
+        // Get all tracking points, or an empty array if none.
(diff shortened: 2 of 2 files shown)


modules/Attendance/Filters/AttendanceFilter.php
```diff
@@ -25,12 +25,12 @@ public function status($status)
         return $this->where('status', $status);
     }
 
-    public function start_date($date)
+    public function startDate($date)
```
mabou7agar, line 28: The method name change from `start_date` to `startDate` breaks backward compatibility if this method is being called elsewhere in the codebase. Consider keeping the old method name or adding a deprecation notice.

modules/Attendance/Filters/AttendanceFilter.php
```diff
@@ -25,12 +25,12 @@
...
     {
         return $this->whereDate('clock_in_time', '>=', $date);
     }
 
-    public function end_date($date)
+    public function endDate($date)
```
mabou7agar, line 33: Similarly, changing `end_date` to `endDate` may break existing code that calls this method. Either maintain the original method name or implement proper versioning/deprecation strategy.

modules/Attendance/Presenters/AttendanceTeamPresenter.php
```diff
@@ -33,6 +38,20 @@ public function present(bool $isListing = false): array
...
+            'latest_location' => $latestPoint ? [
+                'latitude'  => (float) $latestPoint['latitude'],
+                'longitude' => (float) $latestPoint['longitude'],
+                'timestamp' => $latestPoint['timestamp'],
+                'accuracy'  => (float) $latestPoint['accuracy'],
+            ] : [
+                'latitude'  => $this->attendance->clock_in_location['latitude'],
+                'longitude' => $this->attendance->clock_in_location['longitude'],
+                'timestamp' => $this->attendance->clock_in_time->format('Y-m-d H:i:s'),
+                 'accuracy'  => 10,
```
mabou7agar, line 52: The hardcoded accuracy value `10` seems arbitrary. Consider using a configurable value or deriving it from the clock_in_location if available.

PR #808, opened 2025-07-27
change consdition of get sub

AmrSaleh2234:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/modules/RoleAndPermission/Repositories/PermissionRepository.php
+++ b/modules/RoleAndPermission/Repositories/PermissionRepository.php
@@ -64,7 +64,7 @@ class PermissionRepository extends BaseRepository
     }
 
     /**
-     * Get permissions filtered by subEntity names where subEntity matches 
+     * Get permissions filtered by subEntity names where subEntity matches
      * the second part of permission name pattern "programName.subEntity.action"
      */
     public function getPermissionsBySubEntities(array $subEntities): Collection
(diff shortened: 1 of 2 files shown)


modules/RoleAndPermission/Repositories/PermissionRepository.php
```diff
@@ -76,7 +76,8 @@ public function getPermissionsBySubEntities(array $subEntities): Collection
         return $this->model->where(function ($query) use ($subEntities) {
             foreach ($subEntities as $subEntity) {
                 // Use REGEXP to match exactly the second segment: "anything.subEntity.anything"
-                $query->orWhere('name', 'REGEXP', "^[^.]+\\.{$subEntity}\\.[^.]+");
+                $query->orWhere('name', 'LIKE', ".".$subEntity.".");
```
mabou7agar, line 79: The change from REGEXP to LIKE significantly alters the matching logic. The REGEXP was ensuring exact matches of the middle segment (`subEntity`), while LIKE with `.%subEntity%.` will match any occurrence of the subEntity in the name, which could lead to false positives. For example, it would match both `module.subEntity.action` and `module.subEntityX.action`. Consider if this is the intended behavior.

PR #809, opened 2025-07-27
change consdition of get sub

AmrSaleh2234:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/modules/RoleAndPermission/Repositories/PermissionRepository.php
+++ b/modules/RoleAndPermission/Repositories/PermissionRepository.php
@@ -76,7 +76,7 @@ class PermissionRepository extends BaseRepository
         return $this->model->where(function ($query) use ($subEntities) {
             foreach ($subEntities as $subEntity) {
                 // Use REGEXP to match exactly the second segment: "anything.subEntity.anything"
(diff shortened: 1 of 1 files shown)


modules/RoleAndPermission/Repositories/PermissionRepository.php
```diff
@@ -76,7 +76,7 @@ public function getPermissionsBySubEntities(array $subEntities): Collection
         return $this->model->where(function ($query) use ($subEntities) {
             foreach ($subEntities as $subEntity) {
                 // Use REGEXP to match exactly the second segment: "anything.subEntity.anything"
-                $query->orWhere('name', 'LIKE', ".".$subEntity.".");
+                $query->orWhere('name', 'LIKE', "%".".".$subEntity."."."%");
```
mabou7agar, line 79: The change from `LIKE "%.".$subEntity."."%"` to `LIKE "%.".$subEntity."."%"` doesn't actually modify the behavior since both formats will match the same patterns. If the intention was to make the pattern more flexible, consider using REGEXP as mentioned in the comment above for more precise matching.

PR #817, opened 2025-07-27
add LiveTrackingPresenter

AbdeleahmanEl-Desoky:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/modules/Attendance/Presenters/ConstraintPresenter.php
+++ b/modules/Attendance/Presenters/ConstraintPresenter.php
@@ -19,6 +19,7 @@ class ConstraintPresenter extends AbstractPresenter
             'id' => (string) $this->constraint->id,
             'constraint_name' => $this->constraint->constraint_name,
             'constraint_type' =>  __('validation.'.$this->constraint->constraint_type),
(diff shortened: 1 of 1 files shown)


modules/Attendance/Presenters/ConstraintPresenter.php
```diff
@@ -19,6 +19,7 @@ public function present(bool $isListing = false): array
             'id' => (string) $this->constraint->id,
             'constraint_name' => $this->constraint->constraint_name,
             'constraint_type' =>  __('validation.'.$this->constraint->constraint_type),
+            'constraint_code' =>  $this->constraint->constraint_type,
```
mabou7agar, line 22: The added line introduces inconsistency in the data format. The `constraint_type` is translated (line 21) but `constraint_code` uses the raw value. Consider either:
1. Using the translated value for consistency
2. Renaming to `raw_constraint_type` to make the purpose clearer
3. Adding documentation about the difference between these fields if the distinction is intentional

PR #821, opened 2025-07-27
.

AmrSaleh2234:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/modules/RoleAndPermission/Repositories/PermissionRepository.php
+++ b/modules/RoleAndPermission/Repositories/PermissionRepository.php
@@ -76,9 +76,9 @@ class PermissionRepository extends BaseRepository
         return $this->model->where(function ($query) use ($subEntities) {
             foreach ($subEntities as $subEntity) {
                 // Escape special regex characters in subEntity to treat them as literals
-                $escapedSubEntity = preg_quote($subEntity, '/');
+//                $escapedSubEntity = preg_quote($subEntity, '/');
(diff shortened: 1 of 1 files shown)


modules/RoleAndPermission/Repositories/PermissionRepository.php
```diff
@@ -76,9 +76,9 @@ public function getPermissionsBySubEntities(array $subEntities)
         return $this->model->where(function ($query) use ($subEntities) {
             foreach ($subEntities as $subEntity) {
                 // Escape special regex characters in subEntity to treat them as literals
-                $escapedSubEntity = preg_quote($subEntity, '/');
+//                $escapedSubEntity = preg_quote($subEntity, '/');
                 // Use REGEXP to match exactly the second segment: "anything.subEntity.anything"
-                $query->orWhere('name', 'REGEXP', "^[^.]+\\.{$escapedSubEntity}\\.[^.]+");
+                $query->orWhere('name', 'LIKE', "%.".$subEntity.".%");
```
mabou7agar, line 81: The change from REGEXP to LIKE with wildcards might lead to false positives in matching. For example, a subEntity `user` would also match `useradmin`. Consider using exact matching with `CONCAT('%', ?, '%')` or keeping the REGEXP approach if precise matching is required.

PR #822, opened 2025-07-27
add LiveTrackingPresenter

AbdeleahmanEl-Desoky:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/modules/Attendance/Services/AttendanceService.php
+++ b/modules/Attendance/Services/AttendanceService.php
@@ -4,21 +4,15 @@ declare(strict_types=1);
 
 namespace Modules\Attendance\Services;
 
-use BasePackage\Shared\Services\AbstractService;
 use Carbon\Carbon;
-use Illuminate\Pagination\LengthAwarePaginator;
 use Illuminate\Support\Collection;
-use Illuminate\Support\Facades\DB;
 use Modules\Attendance\Models\Attendance;
 use Modules\Attendance\Models\AttendanceBreak;
-use Modules\Attendance\Presenters\AttendancePresenter;
 use Modules\Attendance\Repositories\AttendanceRepository;
 use Modules\Attendance\Exceptions\AttendanceException;
 use Modules\Attendance\DTO\ClockInDTO;
 use Modules\Attendance\DTO\ClockOutDTO;
-use Modules\Attendance\DTO\BreakDTO;
 use Modules\User\Models\User;
-use Modules\User\Models\CompanyUser;
 use Ramsey\Uuid\Uuid;
 use Ramsey\Uuid\UuidInterface;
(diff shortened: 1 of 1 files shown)


modules/Attendance/Services/AttendanceService.php
```diff
@@ -34,23 +28,36 @@ public function __construct(
...
     public function clockIn(ClockInDTO $clockInDTO): Attendance
     {
-        // Check if user is already clocked in
         $existingAttendance = $this->attendanceRepository->getCurrentAttendance($clockInDTO->getUserId());
 
         if ($existingAttendance && !$existingAttendance->clock_out_time) {
             throw AttendanceException::alreadyClockedIn();
         }
-
+        $user = User::find(auth()->user()->id);
```
mabou7agar, line 36: Instead of using `auth()->user()->id` directly, consider passing the authenticated user ID through the DTO or constructor to make the dependency explicit and improve testability.

modules/Attendance/Services/AttendanceService.php
```diff
@@ -34,23 +28,36 @@
...
+            $date = $constraints['next_work_period']['date'];
+            $day_status = $constraints['day_status'];
+        }
         // Create new attendance record
         $attendanceData = [
             'user_id' => $clockInDTO->getUserId(),
             'company_id' => $clockInDTO->getCompanyId(),
             'clock_in_time' => $clockInDTO->getClockInTime(),
             'clock_in_location' => $clockInDTO->getLocation(),
+            'start_time' => $date .' '.$periodStartTime,
```
mabou7agar, line 54: The `$date` variable might be undefined if constraints don't exist. Add a null check before using it to avoid potential undefined variable errors.

modules/Attendance/Services/AttendanceService.php
```diff
@@ -34,23 +28,36 @@
...
+            'end_time' => $date.' '.$periodEndTime,
```
mabou7agar, line 55: Similar to line 54, `$periodEndTime` might be undefined. Ensure proper null handling for these time-related fields.

modules/Attendance/Services/AttendanceService.php
```diff
@@ -34,23 +28,36 @@
...
             'notes' => $clockInDTO->getNotes(),
             'ip_address' => $clockInDTO->getIpAddress(),
             'user_agent' => $clockInDTO->getUserAgent(),
             'status' => 'active',
+            'day_status' => $day_status,
```
mabou7agar, line 60: The `day_status` field is being set without any validation. Consider validating this value against expected statuses if there are specific allowed values.

PR #823, opened 2025-07-27
.

AmrSaleh2234:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/modules/RoleAndPermission/Repositories/PermissionRepository.php
+++ b/modules/RoleAndPermission/Repositories/PermissionRepository.php
@@ -78,7 +78,7 @@ class PermissionRepository extends BaseRepository
                 // Escape special regex characters in subEntity to treat them as literals
 //                $escapedSubEntity = preg_quote($subEntity, '/');
                 // Use REGEXP to match exactly the second segment: "anything.subEntity.anything"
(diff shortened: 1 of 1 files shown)


modules/RoleAndPermission/Repositories/PermissionRepository.php
```diff
@@ -78,7 +78,7 @@ public function getPermissionsBySubEntities(array $subEntities)
                 // Escape special regex characters in subEntity to treat them as literals
 //                $escapedSubEntity = preg_quote($subEntity, '/');
                 // Use REGEXP to match exactly the second segment: "anything.subEntity.anything"
-                $query->orWhere('name', 'LIKE', "%.".$subEntity.".%");
+                $query->orWhere('name', 'LIKE', "%.".$subEntity."%");
```
mabou7agar, line 81: The change from `%.".$subEntity.".%` to `%.".$subEntity."%` might lead to unintended matches. For example, a permission named `module.subEntityExtra` would now match when searching for `subEntity`. Consider if this is the intended behavior or if the original pattern was more precise.

PR #824, opened 2025-07-27
add LiveTrackingPresenter

AbdeleahmanEl-Desoky:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/modules/Attendance/Services/AttendanceConstraintService.php
+++ b/modules/Attendance/Services/AttendanceConstraintService.php
@@ -111,6 +111,12 @@ class AttendanceConstraintService
      */
     public function getEffectiveConstraintForUser(User $user)
     {
+        $constraints = [];
+        $constraint = $user->professionalData?->attendance_constraint;
+        if ($constraint) {
+            $constraints[] = $constraint;
+            return $constraints;
+        }
         $userBranch = $user->userProfessionalData?->branch;
         $userBranchId = $userBranch ? (string) $userBranch->id : null;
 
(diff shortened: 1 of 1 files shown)


modules/Attendance/Services/AttendanceConstraintService.php
```diff
@@ -111,6 +111,12 @@ public function validateAttendance(Attendance $attendance, array $requestData =
      */
     public function getEffectiveConstraintForUser(User $user)
     {
+        $constraints = [];
+        $constraint = $user->professionalData?->attendance_constraint;
+        if ($constraint) {
+            $constraints[] = $constraint;
+            return $constraints;
```
mabou7agar, line 118: The `return $constraints;` statement here will exit the function early and skip the branch constraint logic below. This might not be the intended behavior. Consider removing this return statement if you want to check both user constraints and branch constraints.

modules/Attendance/Services/AttendanceConstraintService.php
```diff
@@ -111,6 +111,12 @@ public function validateAttendance(Attendance $attendance, array $requestData =
      */
     public function getEffectiveConstraintForUser(User $user)
     {
+        $constraints = [];
```
mabou7agar, line 114: The variable `$constraints` is initialized but never used if no constraint is found. This could lead to confusion. Either initialize it only when needed or ensure it's always used.

PR #828, opened 2025-07-27
getEffectiveConstraintForUser

AbdeleahmanEl-Desoky:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/modules/Attendance/Services/AttendanceConstraintService.php
+++ b/modules/Attendance/Services/AttendanceConstraintService.php
@@ -115,7 +115,7 @@ class AttendanceConstraintService
         $constraint = $user->professionalData?->attendanceConstraint;
         if ($constraint) {
             $constraints[] = $constraint;
-            return $constraints;
+            return collect($constraints); 
         }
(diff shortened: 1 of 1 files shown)


modules/Attendance/Services/AttendanceConstraintService.php
```diff
@@ -115,7 +115,7 @@ public function getEffectiveConstraintForUser(User $user)
         $constraint = $user->professionalData?->attendanceConstraint;
         if ($constraint) {
             $constraints[] = $constraint;
-            return $constraints;
+            return collect($constraints);
```
mabou7agar, line 118: The change from returning an array to returning a collection seems unnecessary since the method is expected to return an array (based on the method name and context). This could cause issues with code that expects an array return type. Either:
1. Keep the original array return
2. Update the method signature and all calling code to handle collections
3. Add type hints to clarify the expected return type

PR #838, opened 2025-07-28
change hieracy permissions

AmrSaleh2234:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change


modules/UserInfo/UserProfessionalData/Controllers/UserProfessionalDataController.php
```diff
@@ -33,6 +33,7 @@ public function index(GetUserProfessionalDataListRequest $request): JsonResponse
     {
         $userId = Uuid::fromString($request->route('id'));
         $user = $this->userRepository->getUser($userId);
+        return $user;
```
mabou7agar, line 36: The method `index` is expected to return a `JsonResponse` but is returning the user object directly. This will cause an error since the return type doesn't match. Either remove this line or properly format the response.

PR #847, opened 2025-07-28
.

AmrSaleh2234:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change


modules/Subscription/CompanyAccessProgram/Presenters/CompanyAccessProgramPresenter.php
```diff
@@ -89,7 +89,7 @@ private function calculateProgramsCount(CompanyAccessProgram $companyAccessProgr
         $programsCount = $companyAccessProgram->programs_count ?? 0;
         $subEntitiesCount = $companyAccessProgram->sub_entities_count ?? 0;
 
-        return $programsCount + $subEntitiesCount;
+        return  $subEntitiesCount;
```
mabou7agar, line 92: The change removes the addition of `$programsCount` which seems intentional. However, please confirm if this is the desired behavior as it significantly alters the calculation logic. If this is correct, consider updating the method name `calculateProgramsCount` to better reflect what it's now calculating (e.g., `calculateSubEntitiesCount`).

PR #851, opened 2025-07-28
.

AmrSaleh2234:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/app/Http/Middleware/PermissionMiddleware.php
+++ b/app/Http/Middleware/PermissionMiddleware.php
@@ -23,10 +23,10 @@ class PermissionMiddleware extends SpatiePermissionMiddleware
      *
      * @param  \Illuminate\Http\Request  $request
      * @param  \Closure  $next
-     * @param  string|array  ...$permissions
+     * @param  string|array  $permission
      * @return mixed
      */
-    public function handle($request, Closure $next, ...$permissions)
+    public function handle($request, Closure $next, $permission, $guard = null)
     {
         // Set company ID for multi-tenant environments
         if (!empty(auth('api')->user())) {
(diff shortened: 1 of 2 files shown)


app/Http/Middleware/PermissionMiddleware.php
```diff
@@ -23,10 +23,10 @@ public function __construct(
      *
      * @param  \Illuminate\Http\Request  $request
      * @param  \Closure  $next
-     * @param  string|array  ...$permissions
+     * @param  string|array  $permission
      * @return mixed
      */
-    public function handle($request, Closure $next, ...$permissions)
+    public function handle($request, Closure $next, $permission, $guard = null)
```
mabou7agar, line 29: The method signature change from variadic `...$permissions` to `$permission` is a breaking change that could affect existing route middleware usage. If this is intentional, it should be clearly documented in the PR description under 'Breaking change'.

app/Http/Middleware/PermissionMiddleware.php
```diff
@@ -43,7 +43,7 @@ public function handle($request, Closure $next, ...$permissions)
 //            return $next($request);
 //        }
 
-        $permissions = is_array($permissions[0]) ? $permissions[0] : $permissions;
+        $permissions = explode('|', $permission);
```
mabou7agar, line 46: The variable `$permission` is undefined in this context. It should be using the `$permissions` parameter that's passed to the method. Also, the change from array handling to string splitting with `explode('|', ...)` is a significant behavior change that needs to be properly documented in the PR description.

PR #854, opened 2025-07-29
Feature/attendance system

AbdeleahmanEl-Desoky:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change


modules/Attendance/Resources/routes/api.php
```diff
@@ -65,7 +65,7 @@
         ->name('attendance.team.show');
 
     // Attendance Management (HR/Admin)
-    Route::middleware('permission:manage-attendance')->group(function () {
+    // Route::middleware('permission:manage-attendance')->group(function () {
```
mabou7agar, line 68: The middleware is being commented out but there's no explanation or alternative security measure provided. If this is intentional, consider either:
1. Adding a different middleware for authorization
2. Documenting why the permission check is being removed in the PR description

Security checks shouldn't be removed without proper justification.

modules/Attendance/Resources/routes/api.php
```diff
@@ -77,7 +77,7 @@
 
         Route::delete('{attendanceId}', [AttendanceController::class, 'destroy'])
             ->name('attendance.destroy');
-    });
+    // });
```
mabou7agar, line 80: This appears to be an accidental comment out of a route group closure. Either uncomment it or remove the line completely if it's no longer needed.

PR #856, opened 2025-07-29
fix getTeamAttendance

AbdeleahmanEl-Desoky:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/modules/Attendance/Services/AttendanceService.php
+++ b/modules/Attendance/Services/AttendanceService.php
@@ -473,12 +473,9 @@ public function getTeamAttendance(string $requestingUserId, array $filters, ?int
     // Get all relevant users with optimized eager loading
     $allRelevantUsers = User::where('company_id', $companyId)
         ->filter($filters)
-        ->select(['id', 'name', 'company_id'])
-        ->with([
-            'companyUser:id,user_id,gender,phone,country_id',
-            'companyUser.country:id,name',
-            'professionalData'
-        ])
+        ->select('*')
+        ->with(['professionalData','companyUser'])
+        ->whereHas('companyUser')
         ->get();
 
     $allRelevantUserIds = $allRelevantUsers->pluck('id')->toArray();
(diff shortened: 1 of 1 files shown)


modules/Attendance/Services/AttendanceService.php
```diff
@@ -473,12 +473,9 @@ public function getTeamAttendance(string $requestingUserId, array $filters, ?int
     // Get all relevant users with optimized eager loading
     $allRelevantUsers = User::where('company_id', $companyId)
         ->filter($filters)
-        ->select(['id', 'name', 'company_id'])
-        ->with([
-            'companyUser:id,user_id,gender,phone,country_id',
-            'companyUser.country:id,name',
-            'professionalData'
-        ])
+        ->select('*')
```
mabou7agar, line 476: Changing from selecting specific columns (`id`, `name`, `company_id`) to selecting all columns (`*`) could negatively impact performance, especially with large datasets. Consider keeping the specific column selection for better query efficiency.

modules/Attendance/Services/AttendanceService.php
```diff
@@ -473,12 +473,9 @@
...
+        ->with(['professionalData','companyUser'])
```
mabou7agar, line 477: The eager loading relationship `companyUser.country` was removed which might be needed later in the code. Make sure this removal doesn't break any functionality that depends on the country data.

modules/Attendance/Services/AttendanceService.php
```diff
@@ -473,12 +473,9 @@
...
+        ->whereHas('companyUser')
```
mabou7agar, line 478: The `whereHas('companyUser')` condition is added but it's not clear from the context if this is intentional filtering. Ensure this aligns with the business requirements for team attendance.

modules/Attendance/Services/AttendanceService.php
```diff
@@ -491,8 +488,8 @@
             'user:id,name,email,company_id,status',
         ])
         ->whereIn('user_id', $allRelevantUserIds)
-        ->whereBetween('clock_in_time', [$startDate->startOfDay(), $endDate->endOfDay()])
-        ->orderBy('clock_in_time')
+        ->whereBetween('start_time', [$startDate->startOfDay(), $endDate->endOfDay()])
```
mabou7agar, line 491: The field `start_time` is being used here, but it's not clear if this is the correct field to use for attendance filtering. Please verify that `start_time` accurately represents the attendance clock-in time across the system.

modules/Attendance/Services/AttendanceService.php
```diff
@@ -491,8 +488,8 @@
...
+        ->orderBy('start_time')
```
mabou7agar, line 492: Similarly, ordering by `start_time` might not be equivalent to ordering by `clock_in_time`. Ensure this change doesn't affect the expected sorting behavior of attendance records.

PR #857, opened 2025-07-29
.

AmrSaleh2234:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/modules/RoleAndPermission/Resources/routes/api.php
+++ b/modules/RoleAndPermission/Resources/routes/api.php
@@ -18,7 +18,7 @@ Route::group(['middleware' => ['auth:api', \Stancl\Tenancy\Middleware\Initialize
     // Permission Hierarchy Routes
     Route::group(['prefix' => 'permissions/hierarchy'], function () {
         Route::get('/from-names', [PermissionHierarchyController::class, 'getPermissionsFromNames'])->permission(Permission::PERMISSION_LIST());
-        Route::get('/detailed', [PermissionHierarchyController::class, 'getDetailedPermissions'])->permission(Permission::PERMISSION_LIST());
+        Route::get('/detailed', [PermissionHierarchyController::class, 'getDetailedPermissions']);
     });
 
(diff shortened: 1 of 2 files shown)


modules/RoleAndPermission/Resources/routes/api.php
```diff
@@ -18,7 +18,7 @@
     // Permission Hierarchy Routes
     Route::group(['prefix' => 'permissions/hierarchy'], function () {
         Route::get('/from-names', [PermissionHierarchyController::class, 'getPermissionsFromNames'])->permission(Permission::PERMISSION_LIST());
-        Route::get('/detailed', [PermissionHierarchyController::class, 'getDetailedPermissions'])->permission(Permission::PERMISSION_LIST());
+        Route::get('/detailed', [PermissionHierarchyController::class, 'getDetailedPermissions']);
```
mabou7agar, line 21: Removing the permission check (`Permission::PERMISSION_LIST()`) from this route might be a security concern. Please verify if this endpoint should indeed be publicly accessible without any permission checks.

modules/Subscription/CompanyAccessProgram/Listeners/UpdateMainPackageListener.php
```diff
@@ -224,7 +228,13 @@ private function subEntitiesChanged($companyAccessProgram, array $originalData):
         
         Log::debug("Sub-entities changed result", [
             'company_access_program_id' => $companyAccessProgram->id,
-            'sub_entities_changed' => $subEntitiesChanged
+            'sub_entities_changed' => $subEntitiesChanged,
+            'comparison_details' => [
```
mabou7agar, line 232: The debug log is now including sensitive data (entity IDs) which might not be ideal for production logs. Consider either:
- Removing this debug log entirely if not needed
- Adding a condition to only log this in development/staging environments
- Hashing/obfuscating the IDs before logging

PR #863, opened 2025-07-29

AmrSaleh2234:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/modules/SubEntity/Models/SubEntity.php
+++ b/modules/SubEntity/Models/SubEntity.php
@@ -180,7 +180,7 @@ class SubEntity extends Model
         foreach (self::PERMISSION_ACTIONS as $action) {
             Permission::firstOrCreate([
                 'name' => "{$module}.{$resource}.{$action}",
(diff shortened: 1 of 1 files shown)


modules/SubEntity/Models/SubEntity.php
```diff
@@ -180,7 +180,7 @@ protected function createDefaultPermissions(): void
         foreach (self::PERMISSION_ACTIONS as $action) {
             Permission::firstOrCreate([
                 'name' => "{$module}.{$resource}.{$action}",
-                "key" => "{$module}.{$resource}.{$action}",
+                "key" => "dynamic-".$this->slug.".$action",
```
mabou7agar, line 183: The key format change from `{$module}.{$resource}.{$action}` to `dynamic-{$this->slug}.{$action}` seems inconsistent. Ensure this change won't break existing permission checks elsewhere in the system that might rely on the original format.

PR #888, opened 2025-07-30
handel all sub entity

AmrSaleh2234:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change


modules/Subscription/CompanyAccessProgram/Providers/CompanyAccessProgramServiceProvider.php
```diff
@@ -39,6 +42,14 @@ public function mapRoutes(): void
             ->group($this->getModulePath() . '/Resources/routes/api.php');
     }
 
+    /**
+     * Register model observers
+     */
+    private function registerObservers(): void
```
mabou7agar, line 48: The method `registerObservers` is declared as private but it's not called anywhere in the code. Either make it public and call it in the appropriate place or remove it if not needed.

PR #891, opened 2025-07-30
..

AmrSaleh2234:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change


modules/Subscription/CompanyAccessProgram/Observers/CompanyAccessProgramObserver.php
```diff
@@ -91,22 +91,22 @@ public function deleting(CompanyAccessProgram $companyAccessProgram): ?bool
     public function updating(CompanyAccessProgram $companyAccessProgram): ?bool
     {
         // Check if this is a main program
-        if ($companyAccessProgram->is_main_program) {
+        if (!$companyAccessProgram->is_main_program) {
```
mabou7agar, line 94: The logic inversion (`!$companyAccessProgram->is_main_program`) seems incorrect. The original code was checking if it IS a main program to prevent updates, now it's checking if it's NOT a main program. This would allow updates to main programs and block updates to non-main programs, which contradicts the error message and log warning.

PR #892, opened 2025-07-30
,,

AmrSaleh2234:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change


modules/Subscription/CompanyAccessProgram/Observers/CompanyAccessProgramObserver.php
```diff
@@ -91,7 +91,7 @@ public function deleting(CompanyAccessProgram $companyAccessProgram): ?bool
     public function updating(CompanyAccessProgram $companyAccessProgram): ?bool
     {
         // Check if this is a main program
-        if (!$companyAccessProgram->is_main_program) {
+        if ($companyAccessProgram->is_main_program) {
```
mabou7agar, line 94: The logic inversion here changes the behavior significantly. Please confirm this is intentional as it will now only log when updating a main program instead of non-main programs. This might affect monitoring and error tracking.

PR #894, opened 2025-07-30
Feature/attendance system

AbdeleahmanEl-Desoky:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/modules/Attendance/Presenters/AttendancePresenter.php
+++ b/modules/Attendance/Presenters/AttendancePresenter.php
@@ -33,6 +33,9 @@ class AttendancePresenter extends AbstractPresenter
             'clock_out_time' => $this->attendance->clock_out_time?->setTimezone(
                 new \DateTimeZone($this->attendance->timezone ?? config('app.timezone'))
             )->format('Y-m-d H:i:s'),
+            'start_time' => $this->attendance->start_time,
+            'end_time' => $this->attendance->end_time,
+
 
             'timezone' => $this->attendance->timezone ?? null,
 
--- a/modules/Attendance/Services/TimeConstraintService.php
+++ b/modules/Attendance/Services/TimeConstraintService.php
@@ -531,6 +531,7 @@ class TimeConstraintService extends BaseConstraintService implements TimeConstra
             $periodEnd = Carbon::createFromFormat('H:i', $period['end_time'], $clockInTime->timezone)
                 ->setDateFrom($clockInTime);
 
+
             if ($periodEnd->lessThan($periodStart)) {
                 $periodEnd->addDay();
             }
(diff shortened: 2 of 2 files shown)


modules/Attendance/Presenters/AttendancePresenter.php
```diff
@@ -33,6 +33,9 @@ public function present(bool $isListing = false): array
             'clock_out_time' => $this->attendance->clock_out_time?->setTimezone(
                 new \DateTimeZone($this->attendance->timezone ?? config('app.timezone'))
             )->format('Y-m-d H:i:s'),
+            'start_time' => $this->attendance->start_time,
```
mabou7agar, line 36: Consider formatting `start_time` similar to lines 33-35 for consistency in date/time presentation.

modules/Attendance/Presenters/AttendancePresenter.php
```diff
@@ -33,6 +33,9 @@
...
+            'end_time' => $this->attendance->end_time,
```
mabou7agar, line 37: Consider formatting `end_time` similar to lines 33-35 for consistency in date/time presentation.

modules/Attendance/Services/TimeConstraintService.php
```diff
@@ -531,6 +531,7 @@ public function validateMultiplePeriods(Attendance $attendance, array $config):
             $periodEnd = Carbon::createFromFormat('H:i', $period['end_time'], $clockInTime->timezone)
                 ->setDateFrom($clockInTime);
 
+
```
mabou7agar, line 534: Remove unnecessary blank line with `+` character as it doesn't add any value to the code.

modules/Attendance/Services/TimeConstraintService.php
```diff
@@ -557,13 +558,25 @@
...
-            // تحقق من أن وقت الدخول داخل أي فترة (بما في ذلك سماحية grace)
+
             if ($this->isTimeWithinRangeWithGrace($clockInTime->format('H:i'), $period)) {
                 $inAllowedPeriod = true;
                 break;
             }
         }
-
+        
+        if($period['end_time'] < $clockInTime->format('H:i')){
```
mabou7agar, line 568: The new condition `if($period['end_time'] < $clockInTime->format('H:i'))` is checking each period individually, but it should be checking after all periods have been processed (outside the loop). The current placement might cause false positives since it could return early before checking all periods.

mabou7agar, line 568: The condition should use strict comparison (`===` or `!==`) when comparing times to avoid potential type coercion issues.

modules/Attendance/Services/TimeConstraintService.php
```diff
@@ -557,13 +558,25 @@ public function validateMultiplePeriods(Attendance $attendance, array $config):
...
+            return [
```
mabou7agar, line 569: The error message 'Clock-in time is outside of all allowed work periods for this day' is duplicated in both the new condition and the existing one at line 581. Consider consolidating these checks to avoid duplicate messages.

PR #899, opened 2025-07-30
make checkLateness

AbdeleahmanEl-Desoky:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/modules/Attendance/Services/TimeConstraintService.php
+++ b/modules/Attendance/Services/TimeConstraintService.php
@@ -526,10 +526,42 @@ class TimeConstraintService extends BaseConstraintService implements TimeConstra
                 continue;
             }
 
-            $periodStart = Carbon::createFromFormat('H:i', $period['start_time'], $clockInTime->timezone)
-                ->setDateFrom($clockInTime);
-            $periodEnd = Carbon::createFromFormat('H:i', $period['end_time'], $clockInTime->timezone)
-                ->setDateFrom($clockInTime);
+            // Ensure period times are properly formatted before creating Carbon objects
+            $startTime = trim($period['start_time']);
+            $endTime = trim($period['end_time']);
+            
+            // Validate time format
+            if (!preg_match('/^([0-1]?[0-9]|2[0-3]):[0-5][0-9]$/', $startTime)) {
+                \Illuminate\Support\Facades\Log::error('Invalid period start time format', [
+                    'start_time' => $startTime,
+                    'period' => $period,
+                    'day_of_week' => $dayOfWeek
+                ]);
+                continue;
+            }
+            
+            if (!preg_match('/^([0-1]?[0-9]|2[0-3]):[0-5][0-9]$/', $endTime)) {
+                \Illuminate\Support\Facades\Log::error('Invalid period end time format', [
+                    'end_time' => $endTime,
+                    'period' => $period,
+                    'day_of_week' => $dayOfWeek
+                ]);
+                continue;
+            }
+            
+            try {
+                $periodStart = Carbon::createFromFormat('H:i', $startTime, $clockInTime->timezone)
+                    ->setDateFrom($clockInTime);
+                $periodEnd = Carbon::createFromFormat('H:i', $endTime, $clockInTime->timezone)
+                    ->setDateFrom($clockInTime);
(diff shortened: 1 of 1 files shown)


modules/Attendance/Services/TimeConstraintService.php
```diff
@@ -526,10 +526,42 @@ public function validateMultiplePeriods(Attendance $attendance, array $config):
...
-            $periodStart = Carbon::createFromFormat('H:i', $period['start_time'], $clockInTime->timezone)
-                ->setDateFrom($clockInTime);
-            $periodEnd = Carbon::createFromFormat('H:i', $period['end_time'], $clockInTime->timezone)
-                ->setDateFrom($clockInTime);
+            // Ensure period times are properly formatted before creating Carbon objects
+            $startTime = trim($period['start_time']);
+            $endTime = trim($period['end_time']);
+            
+            // Validate time format
+            if (!preg_match('/^([0-1]?[0-9]|2[0-3]):[0-5][0-9]$/', $startTime)) {
```
mabou7agar, line 534: The regex pattern for time validation could be simplified to `'/^\d{2}:\d{2}$/'` since we already know the format should be HH:MM.

modules/Attendance/Services/TimeConstraintService.php
```diff
@@ -526,10 +526,42 @@
...
+                \Illuminate\Support\Facades\Log::error('Invalid period start time format', [
+                    'start_time' => $startTime,
+                    'period' => $period,
+                    'day_of_week' => $dayOfWeek
+                ]);
+                continue;
+            }
+            
+            if (!preg_match('/^([0-1]?[0-9]|2[0-3]):[0-5][0-9]$/', $endTime)) {
```
mabou7agar, line 543: Same as above - the regex pattern for end time validation could be simplified to `'/^\d{2}:\d{2}$/'`.

modules/Attendance/Services/TimeConstraintService.php
```diff
@@ -526,10 +526,42 @@
...
+                \Illuminate\Support\Facades\Log::error('Invalid period end time format', [
+                    'end_time' => $endTime,
+                    'period' => $period,
+                    'day_of_week' => $dayOfWeek
+                ]);
+                continue;
+            }
+            
+            try {
```
mabou7agar, line 552: The try-catch block is redundant since we've already validated the time format above. Carbon's `createFromFormat` will throw an exception only if the format is invalid, which we've already checked.

modules/Attendance/Services/TimeConstraintService.php
```diff
@@ -670,9 +702,31 @@
...
         }
 
-        // Calculate the earliest allowed clock-in time based on the scheduled start time
-        $scheduledStartTime = Carbon::createFromTimeString($firstPeriod['start_time'])
-                                  ->setDateFrom($clockInTimeLocal);
+        // Ensure start time is properly formatted
+        $startTime = trim($firstPeriod['start_time']);
+        
+        // Validate time format
+        if (!preg_match('/^([0-1]?[0-9]|2[0-3]):[0-5][0-9]$/', $startTime)) {
```
mabou7agar, line 709: The regex pattern `/^([0-1]?[0-9]|2[0-3]):[0-5][0-9]$/` doesn't account for seconds in the time format. If seconds might be present in the input, consider updating the pattern to include them.

modules/Attendance/Services/TimeConstraintService.php
```diff
@@ -670,9 +702,31 @@
...
+            \Illuminate\Support\Facades\Log::error('Invalid period start time format in validateEarlyClockInPrevention', [
+                'start_time' => $startTime,
+                'attendance_id' => $attendance->id,
+                'day_of_week' => $dayOfWeek
+            ]);
+            return false;
+        }
+        
+        try {
+            // Calculate the earliest allowed clock-in time based on the scheduled start time
```
mabou7agar, line 719: The error handling for Carbon object creation is good, but consider adding the original `$firstPeriod['start_time']` value to the error log for better debugging context.

PR #928, opened 2025-07-31
assign program except in central company in main package

AmrSaleh2234:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/modules/Subscription/Package/Presenters/PackagePresenter.php
+++ b/modules/Subscription/Package/Presenters/PackagePresenter.php
@@ -22,7 +22,7 @@ class PackagePresenter extends AbstractPresenter
             'id' => $this->package->id,
             'name' => $this->package->name,
             'status' => $this->package->is_active,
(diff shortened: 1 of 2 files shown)


modules/Subscription/Package/Presenters/PackagePresenter.php
```diff
@@ -22,7 +22,7 @@ protected function present(bool $isListing = false): array
             'id' => $this->package->id,
             'name' => $this->package->name,
             'status' => $this->package->is_active,
-            'features_count' => $this->package->features_count ?? 0,
+            'features_count' => $this->package->permissions()->count() ?? 0,
```
mabou7agar, line 25: The change from `features_count` to `permissions()->count()` might have performance implications since it's making a database query for each package. Consider eager loading the permissions if this is used in a loop or consider caching the count if it doesn't change frequently.

PR #932, opened 2025-07-31
update login

AmrSaleh2234:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/modules/Auth/Services/AuthService.php
+++ b/modules/Auth/Services/AuthService.php
@@ -58,8 +58,9 @@ class AuthService
             $this->sendOtpEmail->loginWithOtp($user->id);
             return [null, $user];
         }
+        $user = $this->userRepository->getUserByIdentifier($authDTO->getEmail());
 
-        $token = JWTAuth::attempt($authDTO->toArray());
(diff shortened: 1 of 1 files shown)


modules/Auth/Services/AuthService.php
```diff
@@ -58,8 +58,9 @@ public function login(LoginDTO $authDTO)
             $this->sendOtpEmail->loginWithOtp($user->id);
             return [null, $user];
         }
+        $user = $this->userRepository->getUserByIdentifier($authDTO->getEmail());
 
-        $token = JWTAuth::attempt($authDTO->toArray());
+        $token = JWTAuth::fromUser($user);
```
mabou7agar, line 63: The change from `JWTAuth::attempt()` to `JWTAuth::fromUser()` introduces a potential security issue. The original code validated the credentials (email and password) before generating a token, while the new code bypasses this validation. This could allow unauthorized access if the user object is obtained incorrectly.

PR #938, opened 2025-07-31

AmrSaleh2234:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/app/Http/Middleware/PermissionMiddleware.php
+++ b/app/Http/Middleware/PermissionMiddleware.php
@@ -51,6 +51,9 @@ class PermissionMiddleware extends SpatiePermissionMiddleware
         if(auth()->check() && auth()->user()->email == 'admin@constrix-nv.com') {
             return $next($request);
         }
+        if(auth()->check() && auth()->user()->hasRole('super-admin') && tenant("is_central_company")) {
+            return $next($request);
+        }
 
         $permissions = explode('|', $permission);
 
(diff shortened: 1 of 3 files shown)


app/Http/Middleware/PermissionMiddleware.php
```diff
@@ -51,6 +51,9 @@ public function handle($request, Closure $next, $permission, $guard = null)
         if(auth()->check() && auth()->user()->email == 'admin@constrix-nv.com') {
             return $next($request);
         }
+        if(auth()->check() && auth()->user()->hasRole('super-admin') && tenant("is_central_company")) {
```
mabou7agar, line 54: The condition `tenant("is_central_company")` should be checked before `hasRole('super-admin')` for better performance since checking the tenant flag is likely a simpler operation than checking user roles.

modules/RoleAndPermission/Database/Seeders/RolesAndPermissionsSeeder.php
```diff
@@ -26,7 +26,7 @@ public function run()
         Model::unguard();
 
         // Get current company ID or use the first company
-        $companyId = tenant("id") ?? Company::query()->first()?->id;
+        $companyId = tenant("id") ?? "560005d6-04b8-53b3-9889-d312648288e3";//id form new vision company
```
mabou7agar, line 29: Hardcoding a company ID is not recommended as it makes the code less flexible and could cause issues in different environments. Consider using the original approach of getting the first company's ID or provide a way to configure this ID through environment variables or config files.

PR #942, opened 2025-07-31
..

AmrSaleh2234:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/modules/User/Resources/routes/api.php
+++ b/modules/User/Resources/routes/api.php
@@ -5,7 +5,7 @@ use Modules\User\Controllers\UserController;
 use Modules\RoleAndPermission\Enums\Permission;
 
 Route::group(['middleware' => ['auth:api', \Stancl\Tenancy\Middleware\InitializeTenancyByRequestData::class]], function () {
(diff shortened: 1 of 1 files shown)


modules/User/Resources/routes/api.php
```diff
@@ -5,7 +5,7 @@
 use Modules\RoleAndPermission\Enums\Permission;
 
 Route::group(['middleware' => ['auth:api', \Stancl\Tenancy\Middleware\InitializeTenancyByRequestData::class]], function () {
-    Route::get('/', [UserController::class, 'index'])->name("users.list")->permission(Permission::USER_LIST());
+    Route::get('/', [UserController::class, 'index'])->name("users.list");
```
mabou7agar, line 8: The permission check `->permission(Permission::USER_LIST())` was removed from this route. This might be a security concern if the route should indeed be protected. Please confirm if this was intentional and document the reason if so.

PR #943, opened 2025-07-31
Feature/attendance system

AbdeleahmanEl-Desoky:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/lang/en/validation.php
+++ b/lang/en/validation.php
@@ -519,5 +519,6 @@ return [
         'holiday'=>'holiday',
         'day_off_or_weekend'=> 'day off or weekend',
         'in_loction'=>'in loction',
+        'clocked_out'=> 'clocked out'
     ]
 ];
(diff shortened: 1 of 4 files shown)


lang/en/validation.php
```diff
@@ -519,5 +519,6 @@
         'holiday'=>'holiday',
         'day_off_or_weekend'=> 'day off or weekend',
         'in_loction'=>'in loction',
```
mabou7agar, line 521: There's a typo in the key `in_loction`, it should be `in_location` to match the correct spelling.

modules/Attendance/Services/AttendanceService.php
```diff
@@ -620,7 +620,7 @@ public function getTeamAttendance(string $requestingUserId, array $filters, ?int
                         'day_status' => $isHoliday === true ? 'holiday' : 'work_day',
                         'id' => Uuid::uuid4(),
                         'is_late' => false,
-                        'clock_in_time' => null,
+                        'clock_in_time' => $dateStr,
```
mabou7agar, line 623: The change from `null` to `$dateStr` for `clock_in_time` seems incorrect. `clock_in_time` should either be a valid timestamp when clocking in or remain `null` if not clocked in. Using `$dateStr` (which appears to be just a date string) as the clock-in time doesn't make logical sense.

PR #946, opened 2025-07-31
.

AmrSaleh2234:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/app/Exceptions/Handler.php
+++ b/app/Exceptions/Handler.php
@@ -33,7 +33,7 @@ class Handler
             $e instanceof UnauthorizedException => response()->json([
                 'success' => false,
                 'message' => __('validation.unauthorized'),
-            ], 403),
+            ], 404),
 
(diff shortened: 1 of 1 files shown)


app/Exceptions/Handler.php
```diff
@@ -33,7 +33,7 @@ public static function handle(Throwable $e): JsonResponse
             $e instanceof UnauthorizedException => response()->json([
                 'success' => false,
                 'message' => __('validation.unauthorized'),
-            ], 403),
+            ], 404),
```
mabou7agar, line 36: Changing status code from 403 to 404 for UnauthorizedException seems incorrect semantically. 403 (Forbidden) is the appropriate status code for unauthorized access, while 404 (Not Found) should be used when a resource doesn't exist. Consider reverting this change unless there's a specific reason for it.

PR #947, opened 2025-07-31
.

AmrSaleh2234:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/app/Exceptions/Handler.php
+++ b/app/Exceptions/Handler.php
@@ -33,6 +33,8 @@ class Handler
             $e instanceof UnauthorizedException => response()->json([
                 'success' => false,
                 'message' => __('validation.unauthorized'),
+                'error' => $e->getMessage() , // Hide error details in production
+                "trace"=> $e->getTrace(), // Hide error details in production <==>
             ], 404),
 
             $e instanceof NotFoundHttpException => response()->json([


app/Exceptions/Handler.php
```diff
@@ -33,6 +33,8 @@ public static function handle(Throwable $e): JsonResponse
             $e instanceof UnauthorizedException => response()->json([
                 'success' => false,
                 'message' => __('validation.unauthorized'),
+                'error' => $e->getMessage() , // Hide error details in production
```
mabou7agar, line 36: Exposing error messages and stack traces in production is a security risk. Consider removing these lines or wrapping them in an environment check (e.g., `if (!app()->isProduction())`).

app/Exceptions/Handler.php
```diff
@@ -33,6 +33,8 @@ public static function handle(Throwable $e): JsonResponse
...
+                "trace"=> $e->getTrace(), // Hide error details in production <==>
```
mabou7agar, line 37: Stack traces should never be exposed in production as they may reveal sensitive information about the application's structure and implementation details.

PR #949, opened 2025-07-31

AmrSaleh2234:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/database/seeders/MainPackageSeeder.php
+++ b/database/seeders/MainPackageSeeder.php
@@ -26,13 +26,15 @@ class MainPackageSeeder extends Seeder
     public function run(): void
     {
         DB::transaction(function () {
-            // 1. Create Main Company Access Program
-            $accessProgram = CompanyAccessProgram::firstOrCreate([
-                'name' => 'Main Access Program',
-            ], [
-                'is_active' => true,
-                'is_main_program' => true,
-            ]);
+            // 1. Create Main Company Access Program without triggering observers
+            $accessProgram = CompanyAccessProgram::withoutEvents(function () {
+                return CompanyAccessProgram::firstOrCreate([
+                    'name' => 'Main Access Program',
+                ], [
+                    'is_active' => true,
+                    'is_main_program' => true,
+                ]);
+            });
 
             // 1.1. Always sync all countries, company types, and company fields to Main Access Program
             // This ensures newly added data is included when the seeder runs again
@@ -141,13 +143,7 @@ class MainPackageSeeder extends Seeder
                 'is_main_package' =>1
             ]);
 
-            // 3. Sync all permissions except company, company_access_program, and package related ones
-            $excludedPermissionPatterns = [
-                'companies',
-                'users',
-                'subscription',
-                "program-management",
-            ];
+
 
             $totalPermissions = Permission::count();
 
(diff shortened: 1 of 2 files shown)


database/seeders/MainPackageSeeder.php
```diff
@@ -26,13 +26,15 @@ class MainPackageSeeder extends Seeder
...
         DB::transaction(function () {
-            // 1. Create Main Company Access Program
-            $accessProgram = CompanyAccessProgram::firstOrCreate([
-                'name' => 'Main Access Program',
-            ], [
-                'is_active' => true,
-                'is_main_program' => true,
-            ]);
+            // 1. Create Main Company Access Program without triggering observers
+            $accessProgram = CompanyAccessProgram::withoutEvents(function () {
```
mabou7agar, line 30: The `withoutEvents` wrapper seems unnecessary here since seeders typically run in isolation and don't need to suppress events. This could lead to unexpected behavior if there are important model events that should trigger during seeding.

database/seeders/MainPackageSeeder.php
```diff
@@ -49,6 +51,83 @@ public function run(): void
...
+            $processedSubEntities = [];
+
+            foreach ($allPermissions as $permission) {
+                // Parse permission name pattern: {program}.{sub_entity}.{action}
+                $nameParts = explode('.', $permission->name);
+
+                if (count($nameParts) >= 3) {
+                    $program = $nameParts[0];
+                    $subEntity = $nameParts[1];
+                    if (str_contains($subEntity, "*")) {
```
mabou7agar, line 76: The logic for handling wildcard sub-entities could be simplified. Consider using `str_replace('*', '', $subEntity)` instead of manual string manipulation.

database/seeders/MainPackageSeeder.php
```diff
@@ -49,6 +51,83 @@ public function run(): void
...
+                        $resources = explode('*', $nameParts[1]);
+                        $subEntity = $resources[0];
+                        if (uuid_is_valid($resources[1])) {
```
mabou7agar, line 79: The UUID check seems unnecessary here since we're processing permission names which typically don't contain UUIDs. This check might be redundant.

database/seeders/MainPackageSeeder.php
```diff
@@ -49,6 +51,83 @@ public function run(): void
...
+                            $subEntity = $nameParts[1];
+                        }
+
+                    }
+
+                    // Check if program is not in excluded patterns
+                    $isExcluded = false;
```
mabou7agar, line 86: The exclusion check could be more efficient using `Str::contains()` with an array of patterns instead of looping manually.

database/seeders/MainPackageSeeder.php
```diff
@@ -49,6 +51,83 @@ public function run(): void
...
+                        if (!in_array($program, $processedPrograms)) {
+                            $programsData[] = [
+                                'company_access_program_id' => $accessProgram->id,
+                                'program_id' => $program,
+                            ];
+                            $processedPrograms[] = $program;
+                        }
+
+                        // Add sub-entity if not already processed
+                        $subEntityKey = $subEntity;
```
mabou7agar, line 105: The `$subEntityKey` variable assignment is redundant since it's just a copy of `$subEntity`. Consider using `$subEntity` directly in the check.

PR #1116, opened 2025-09-14
Stage

AmrSaleh2234:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/devops/supervisord.conf
+++ b/devops/supervisord.conf
@@ -43,7 +43,7 @@ stdout_logfile_backups=3
 stopwaitsecs=60
 
 [program:laravel-scheduler]
-command=php /var/www/artisan schedule:work
+command=bash -c "while true; do php /var/www/artisan schedule:run && sleep 60; done"
 directory=/var/www
 autostart=true
 autorestart=true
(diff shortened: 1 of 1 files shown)


devops/supervisord.conf
```diff
@@ -43,7 +43,7 @@ stdout_logfile_backups=3
 stopwaitsecs=60
 
 [program:laravel-scheduler]
-command=php /var/www/artisan schedule:work --sleep=3 --tries=3 --max-time=3600 --memory=512
+command=bash -c "while true; do php /var/www/artisan schedule:run && sleep 60; done"
```
mabou7agar, line 46: The change from `schedule:work` to a bash loop with `schedule:run` introduces potential issues:
- The sleep duration is now fixed at 60 seconds which may not be optimal for all tasks
- No retry mechanism (`--tries=3`) is implemented anymore
- No max execution time (`--max-time=3600`) is enforced
- No memory limit (`--memory=512`) is set

Consider either:
1. Keeping the original `schedule:work` command with its built-in features
2. If the bash loop is necessary, implement proper error handling and resource limits

PR #1128, opened 2025-09-17
update seeder

AmrSaleh2234:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/modules/Program/Database/Seeders/ProgramDatabaseSeeder.php
+++ b/modules/Program/Database/Seeders/ProgramDatabaseSeeder.php
@@ -3,6 +3,7 @@
 namespace Modules\Program\Database\Seeders;
 
 use Illuminate\Database\Seeder;
+use Illuminate\Support\Str;
 use Modules\Program\Models\Program;
 use Illuminate\Database\Eloquent\Model;
 
@@ -27,7 +28,7 @@ class ProgramDatabaseSeeder extends Seeder
         ];
 
         foreach ($programs as $programName) {
-            Program::firstOrCreate (
+            Program::firstOrCreate (["slug"=>Str::slug($programName['en'])],
                 ['name' => $programName]
             );
         }
--- a/modules/Program/Models/Program.php
+++ b/modules/Program/Models/Program.php
@@ -28,7 +28,8 @@ class Program extends Model
     protected $fillable = [
         'name',
         'is_active',
-        'parent_id'
+        'parent_id',
+        "slug"
     ];
 
     protected $casts = [
--- a/modules/SubEntity/Services/SubEntityRecordsService.php
+++ b/modules/SubEntity/Services/SubEntityRecordsService.php
@@ -74,11 +74,11 @@ class SubEntityRecordsService
     {
         // Get filtered query using same logic as the original method but without pagination
         $query = $this->companyUserRepository->getModel();
-        
+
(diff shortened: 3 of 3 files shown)


modules/Program/Database/Seeders/ProgramDatabaseSeeder.php
```diff
@@ -27,7 +28,7 @@ public function run()
         ];
 
         foreach ($programs as $programName) {
-            Program::firstOrCreate (
+            Program::firstOrCreate (["slug"=>Str::slug($programName['en'])],
```
mabou7agar, line 31: The array structure is inconsistent - you're mixing associative and indexed arrays. Either use `['slug' => Str::slug($programName['en']), 'name' => $programName]` or ensure `$programName` is properly formatted as an associative array.

modules/Program/Models/Program.php
```diff
@@ -28,7 +28,8 @@ class Program extends Model
     protected $fillable = [
         'name',
         'is_active',
-        'parent_id'
+        'parent_id',
+        "slug"
```
mabou7agar, line 32: The string delimiter should be consistent with the rest of the file. Change `"` to `'` to match other entries in the array.

modules/SubEntity/Services/SubEntityRecordsService.php
```diff
@@ -95,13 +95,13 @@ protected function getMappedRecordsWidgets($type): array
         })->count();
 
         // Get last month data for comparison
-        $lastMonth = Carbon::now()->subMonth();
+        $lastMonth = Carbon::now();
```
mabou7agar, line 98: Changing `subMonth()` to `now()` here will break the logic since you're trying to get last month's data. This should be reverted to `subMonth()`.

modules/SubEntity/Services/SubEntityRecordsService.php
```diff
@@ -95,13 +95,13 @@ protected function getMappedRecordsWidgets($type): array
...
         $recordsAddedLastMonth = (clone $query)->where('created_at', '>=', $lastMonth->startOfMonth())
                                                ->where('created_at', '<=', $lastMonth->endOfMonth())
                                                ->count();
 
         // Get previous month data for percentage calculation
-        $prevMonth = Carbon::now()->subMonths(2);
+        $prevMonth = Carbon::now()->subMonth();
```
mabou7agar, line 104: Changing from `subMonths(2)` to `subMonth()` changes the logic. If this is intentional, please explain why in the PR description since it affects how the percentage calculation works.

modules/SubEntity/Services/SubEntityRecordsService.php
```diff
@@ -137,13 +137,13 @@ protected function getCustomEntityWidgets(string $model, string $registrationFor
         $suspendedRecords = (clone $query)->where('status', -1)->count();
 
         // Get last month data
-        $lastMonth = Carbon::now()->subMonth();
+        $lastMonth = Carbon::now();
```
mabou7agar, line 140: Changing `subMonth()` to `now()` will make the code calculate records for the current month instead of last month, which contradicts the variable name `$lastMonth` and the comment above it. This appears to be a bug.

PR #1153, opened 2025-09-28
setting get and pust and seeder

AmrSaleh2234:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/modules/NotificationSettings/Repositories/NotificationSettingsRepository.php
+++ b/modules/NotificationSettings/Repositories/NotificationSettingsRepository.php
@@ -32,7 +32,7 @@ class NotificationSettingsRepository extends BaseRepository
 
     public function getNotificationSettings(): NotificationSettings
     {
-        return $this->model->query()->first();
+        return $this->model->query()->where("company_id",tenant("id"))->first();
     }
 
     public function createNotificationSettings(array $data): NotificationSettings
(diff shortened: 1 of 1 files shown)


modules/NotificationSettings/Repositories/NotificationSettingsRepository.php
```diff
@@ -32,7 +32,7 @@ public function getNotificationSettingsList(?int $page, ?int $perPage = 10): Col
 
     public function getNotificationSettings(): NotificationSettings
     {
-        return $this->model->query()->first();
+        return $this->model->query()->where("company_id",tenant("id"))->first();
```
mabou7agar, line 35: The tenant() helper function is being used directly in the query. Consider injecting the tenant ID as a parameter to make the method more testable and decoupled from the tenant helper.

modules/NotificationSettings/Repositories/NotificationSettingsRepository.php
```diff
@@ -42,7 +42,7 @@ public function createNotificationSettings(array $data): NotificationSettings
 
     public function updateNotificationSettings( array $data): bool
     {
-        return $this->model->query()->first()->update( $data);
+        return $this->model->query()->where("company_id",tenant("id"))->first()->update( $data);
```
mabou7agar, line 45: The change adds a company_id filter which is good, but there's no error handling if no record is found. Consider adding a check before calling `update()` to handle cases where no matching record exists.

PR #1167, opened 2025-10-06
Feature/audit

AmrSaleh2234:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/modules/Audit/Controllers/AuditController.php
+++ b/modules/Audit/Controllers/AuditController.php
@@ -37,11 +37,18 @@ class AuditController extends Controller
 
     public function activityLog(GetAuditListRequest $request)
     {
-
         $list = $this->auditService->groupedByDate();
-        return response(["code" => "SUCCESS_WITH_LIST_PAYLOAD_OBJECTS",
+        
+        // Apply presenter to each audit item while maintaining date grouping
+        $formattedList = [];
+        foreach ($list as $date => $audits) {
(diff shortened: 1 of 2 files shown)


modules/Audit/Controllers/AuditController.php
```diff
@@ -37,11 +37,18 @@ public function index(GetAuditListRequest $request)
 
     public function activityLog(GetAuditListRequest $request)
     {
-
         $list = $this->auditService->groupedByDate();
-        return response(["code" => "SUCCESS_WITH_LIST_PAYLOAD_OBJECTS",
+        
+        // Apply presenter to each audit item while maintaining date grouping
```
mabou7agar, line 42: The comment here is unnecessary as the code is self-explanatory. The foreach loop clearly shows it's formatting the list by applying the presenter to each audit item while maintaining the date grouping.

modules/Audit/Controllers/AuditController.php
```diff
@@ -37,11 +37,18 @@ public function index(GetAuditListRequest $request)
...
+        $formattedList = [];
+        foreach ($list as $date => $audits) {
+            $formattedList[$date] = AuditPresenter::collection($audits);
+        }
+        
+        return response([
```
mabou7agar, line 48: The response array formatting has changed but the message is still `null`. Consider providing a meaningful message here instead of null.

PR #1174, opened 2025-10-07
add password

AmrSaleh2234:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/lang/en/validation.php
+++ b/lang/en/validation.php
@@ -408,6 +408,7 @@ return [
     'study_rate_required' => 'Study rate is required.',
     'study_rate_numeric' => 'Study rate must be a number.',
     'graduation_date_required' => 'Graduation date is required.',
+    "access-denied"=>"access denied",
 
   'custom' => [
         'attribute-name' => [
--- a/modules/ArchiveLibrary/Folder/Presenters/FolderPresenter.php
+++ b/modules/ArchiveLibrary/Folder/Presenters/FolderPresenter.php
@@ -27,6 +27,7 @@ class FolderPresenter extends AbstractPresenter
             'file' => $this->folder->getFirstMedia("upload") ? (new MediaPresenter($this->folder->getFirstMedia('upload')))->getData(): null,
             "created_at"=>$this->folder->created_at,
             "updated_at"=>$this->folder->updated_at,
+            "is_password"=>$this->folder->password != null?1 : 0
         ];
     }
 }
--- a/modules/ArchiveLibrary/Folder/Repositories/FolderRepository.php
+++ b/modules/ArchiveLibrary/Folder/Repositories/FolderRepository.php
@@ -8,6 +8,7 @@ use App\Exceptions\CustomException;
 use BasePackage\Shared\Repositories\BaseRepository;
 use Illuminate\Database\Eloquent\Collection;
 use Illuminate\Http\UploadedFile;
+use Illuminate\Support\Facades\Hash;
 use Mockery\Exception;
 use Modules\ArchiveLibrary\File\Models\File;
 use Modules\ArchiveLibrary\File\Models\UserFilePermission;
@@ -24,7 +25,7 @@ use Modules\ArchiveLibrary\Folder\Models\UserFolderPermission;
  */
 class FolderRepository extends BaseRepository
 {
-    public function __construct(Folder $model,private FileUploadService $uploadedFile)
+    public function __construct(Folder $model, private FileUploadService $uploadedFile)
     {
         parent::__construct($model);
     }
@@ -33,7 +34,7 @@ class FolderRepository extends BaseRepository
     {
         $query = $this->model->query();
 
-        if ($parentId !=null) {
+        if ($parentId != null) {
             $query->where('parent_id', $parentId);
         } else {
             $query->whereNull('parent_id');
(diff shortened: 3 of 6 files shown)


lang/en/validation.php
```diff
@@ -408,6 +408,7 @@
     'study_rate_required' => 'Study rate is required.',
     'study_rate_numeric' => 'Study rate must be a number.',
     'graduation_date_required' => 'Graduation date is required.',
+    "access-denied"=>"access denied",
```
mabou7agar, line 411: The string `access-denied` should follow the same naming convention as other keys in this file (snake_case). Suggested change:
```suggestion
    'access_denied' => 'access denied',
```

modules/ArchiveLibrary/Folder/Presenters/FolderPresenter.php
```diff
@@ -27,6 +27,7 @@ protected function present(bool $isListing = false): array
             'file' => $this->folder->getFirstMedia("upload") ? (new MediaPresenter($this->folder->getFirstMedia('upload')))->getData(): null,
             "created_at"=>$this->folder->created_at,
             "updated_at"=>$this->folder->updated_at,
+            "is_password"=>$this->folder->password != null?1 : 0
```
mabou7agar, line 30: Consider using strict comparison (`!== null`) instead of loose comparison (`!= null`) for better type safety.

mabou7agar, line 30: The ternary operator could be simplified to just `(int)($this->folder->password !== null)` since we're converting a boolean to 1/0.

mabou7agar, line 30: The property name `is_password` suggests a boolean value, but we're returning an integer (0/1). Consider either:
1. Keeping it boolean (`true/false`) and changing the property name to `has_password`
2. Or keeping the integer return but renaming to `password_exists`

modules/ArchiveLibrary/Folder/Repositories/FolderRepository.php
```diff
@@ -81,10 +83,9 @@ public function updateFolder(UuidInterface $id, array $data, array $userIds = []
             if (!empty($userIds) || $folder->access_type === 'private') {
                 $folder->users()->sync($userIds);
             }
-            if ($file == null)
-            {
+            if ($file == null) {
                 $folder->clearMediaCollection('upload');
```
mabou7agar, line 87: The logic here seems inverted - you're clearing the media collection when `$file` is null, but then trying to upload a null file. This might be unintended behavior. Shouldn't the media collection be cleared and file uploaded only when `$file` is NOT null?

modules/ArchiveLibrary/Folder/Repositories/FolderRepository.php
```diff
@@ -81,10 +83,9 @@ public function updateFolder(UuidInterface $id, array $data, array $userIds = []
...
-                $this->uploadedFile->uploadFile($folder,$file,'upload');
+                $this->uploadedFile->uploadFile($folder, $file, 'upload');
```
mabou7agar, line 88: There's an extra space after `$folder,` that should be removed for consistency with the rest of the codebase.

modules/ArchiveLibrary/Folder/Repositories/FolderRepository.php
```diff
@@ -144,7 +148,10 @@ public function getFoldersAndFilesByParent(?string $parentId, $userId): array
     {
         // Query folders based on parent_id
         $foldersQuery = $this->model->query();
-
+        $folder = $this->model->query()->where('id', $parentId)->first();
```
mabou7agar, line 151: The password check logic should be moved to a separate method to improve code reusability and maintainability.

modules/ArchiveLibrary/Folder/Repositories/FolderRepository.php
```diff
@@ -144,7 +148,10 @@ public function getFoldersAndFilesByParent(?string $parentId, $userId): array
...
+        if ($folder->password != null  &&( !request()->has("password") || !Hash::check(request()->get("password"), $folder->password))) {
```
mabou7agar, line 152: The condition is hard to read due to multiple nested checks. Consider breaking it down into separate if statements or using helper variables for better readability.

mabou7agar, line 152: Missing null check for `$folder`. If `$parentId` doesn't exist, this will throw an error.

modules/ArchiveLibrary/Folder/Repositories/FolderRepository.php
```diff
@@ -144,7 +148,10 @@ public function getFoldersAndFilesByParent(?string $parentId, $userId): array
...
+             throw new CustomException(__("validation.access-denied"));
```
mabou7agar, line 153: The error message should be more specific about the password requirement rather than just saying 'access denied'.

modules/ArchiveLibrary/Folder/Services/FolderCRUDService.php
```diff
@@ -124,6 +124,7 @@ public function listFolders(UuidInterface $userId, ?string $parentId, int $page
 
     public function getFoldersAndFiles( $userId, ?string $parentId): array
     {
+
```
mabou7agar, line 127: Remove unnecessary empty line added before the return statement

PR #1193, opened 2025-10-09
get size for folder

AmrSaleh2234:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/modules/ArchiveLibrary/Folder/Presenters/FolderPresenter.php
+++ b/modules/ArchiveLibrary/Folder/Presenters/FolderPresenter.php
@@ -86,19 +86,19 @@ class FolderPresenter extends AbstractPresenter
         $folderIds = $folders->pluck('id')->toArray();
 
         // Get all files for all folders with their media sizes in a single optimized query
-        $fileSizes = \Modules\ArchiveLibrary\File\Models\File::whereIn('folder_id', $folderIds)
-            ->with(['media' => function ($query) {
-                $query->select('id', 'model_id', 'size');
-            }])
-            ->get()
-            ->groupBy('folder_id')
-            ->map(function ($files) {
-                // Sum all media sizes for all files in this folder
-                return $files->sum(function ($file) {
-                    return $file->media->sum('size');
-                });
+        // Using DB join for better performance
+        $fileSizes = \DB::table('files')
+            ->join('media', function($join) {
+                $join->on('files.id', '=', 'media.model_id')
+                     ->where('media.model_type', '=', 'Modules\\ArchiveLibrary\\File\\Models\\File')
+                     ->where('media.collection_name', '=', 'upload');
             })
-            ->all();
+            ->whereIn('files.folder_id', $folderIds)
+            ->select('files.folder_id', \DB::raw('SUM(media.size) as total_size'))
+            ->groupBy('files.folder_id')
+            ->get()
+            ->pluck('total_size', 'folder_id')
+            ->toArray();
 
         self::$fileSizesCache = $fileSizes;
     }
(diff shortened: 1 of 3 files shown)


modules/ArchiveLibrary/Folder/Presenters/FolderPresenter.php
```diff
@@ -86,19 +86,19 @@ private static function primeFileSizesCache(Collection $folders): void
...
-                $query->select('id', 'model_id', 'size');
-            }])
-            ->get()
-            ->groupBy('folder_id')
-            ->map(function ($files) {
-                // Sum all media sizes for all files in this folder
-                return $files->sum(function ($file) {
-                    return $file->media->sum('size');
-                });
+        // Using DB join for better performance
```
mabou7agar, line 89: The raw DB query approach is generally more performant, but we should ensure proper escaping of the model type string to prevent SQL injection. Consider using `addslashes()` or prepared statements.

modules/ArchiveLibrary/Folder/Presenters/FolderPresenter.php
```diff
@@ -86,19 +86,19 @@ private static function primeFileSizesCache(Collection $folders): void
...
+        $fileSizes = \DB::table('files')
+            ->join('media', function($join) {
+                $join->on('files.id', '=', 'media.model_id')
+                     ->where('media.model_type', '=', 'Modules\\ArchiveLibrary\\File\\Models\\File')
+                     ->where('media.collection_name', '=', 'upload');
```
mabou7agar, line 94: Hardcoding the collection name 'upload' might cause issues if this collection name changes in the future. Consider making this configurable.

modules/ArchiveLibrary/Folder/Presenters/FolderPresenter.php
```diff
@@ -86,19 +86,19 @@ private static function primeFileSizesCache(Collection $folders): void
...
             })
-            ->all();
+            ->whereIn('files.folder_id', $folderIds)
+            ->select('files.folder_id', \DB::raw('SUM(media.size) as total_size'))
+            ->groupBy('files.folder_id')
+            ->get()
+            ->pluck('total_size', 'folder_id')
+            ->toArray();
```
mabou7agar, line 101: The `toArray()` call might be unnecessary here since `pluck()` with two parameters already returns an associative array. This could be simplified.

modules/ArchiveLibrary/Folder/Presenters/FolderPresenter.php
```diff
@@ -149,18 +149,18 @@ private function getFolderSize(): int
...
-        $totalSize = \Modules\ArchiveLibrary\File\Models\File::where('folder_id', $this->folder->id)
-            ->with(['media' => function ($query) {
-                $query->select('id', 'model_id', 'size');
-            }])
-            ->get()
-            ->sum(function ($file) {
-                return $file->media->sum('size');
-            });
+        $totalSize = \DB::table('files')
+            ->join('media', function($join) {
```
mabou7agar, line 157: The join condition should be moved to a separate line for better readability:
```php
$join->on('files.id', '=', 'media.model_id')
    ->where('media.model_type', '=', 'Modules\\ArchiveLibrary\\File\\Models\\File')
    ->where('media.collection_name', '=', 'upload');
```

modules/ArchiveLibrary/Folder/Presenters/FolderPresenter.php
```diff
@@ -149,18 +149,18 @@ private function getFolderSize(): int
...
+                $join->on('files.id', '=', 'media.model_id')
+                     ->where('media.model_type', '=', 'Modules\\ArchiveLibrary\\File\\Models\\File')
+                     ->where('media.collection_name', '=', 'upload');
+            })
+            ->where('files.folder_id', $this->folder->id)
+            ->sum('media.size');
```
mabou7agar, line 163: Consider adding a comment explaining why we're casting to int here since it's not immediately obvious why this is needed (though I know you said not to suggest adding comments, this seems like an important point that should be addressed).

PR #1203, opened 2025-10-12
.

AmrSaleh2234:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/modules/NotificationSettings/Services/DocumentNotificationService.php
+++ b/modules/NotificationSettings/Services/DocumentNotificationService.php
@@ -173,16 +173,19 @@ class DocumentNotificationService
             $notifyDate = Carbon::parse($item->$dateField);
             
             if ($reminderType === 'daily') {
-                // Send daily notifications for all eligible items
+                // Send daily notifications for all eligible items (due today or overdue)
                 return true;
             }
             
             if ($reminderType === 'weekly') {
-                // Send weekly notifications only on specific day of week or if overdue
-                $daysDifference = $today->diffInDays($notifyDate, false);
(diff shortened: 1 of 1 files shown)


modules/NotificationSettings/Services/DocumentNotificationService.php
```diff
@@ -173,16 +173,19 @@ private function filterItemsByReminderType(Collection $items, string $reminderTy
...
+                // Send weekly notifications every Monday for all overdue/due items
+                // OR send on the exact due date regardless of day
+                $isMonday = $today->dayOfWeek === Carbon::MONDAY;
+                $isDueToday = $today->isSameDay($notifyDate);
+                $isOverdue = $notifyDate->lt($today);
                 
-                // Send if it's exactly the notify date, or if it's Monday and the item is overdue
-                return $daysDifference === 0 || ($today->dayOfWeek === Carbon::MONDAY && $daysDifference < 0);
+                // Send if: it's the exact due date, OR it's Monday and item is due/overdue
+                return $isDueToday || ($isMonday && ($isDueToday || $isOverdue));
```
mabou7agar, line 188: The logic here seems incorrect. `$isDueToday` is already checked in the first condition, so checking it again in the second condition is redundant. Also, the condition `($isMonday && ($isDueToday || $isOverdue))` will never be true when `$isDueToday` is false since it's already checked above. Suggested fix:
```php
return $isDueToday || ($isMonday && $isOverdue);
```

PR #1215, opened 2025-10-14
sync permission

AmrSaleh2234:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/modules/Subscription/Package/Repositories/PackageRepository.php
+++ b/modules/Subscription/Package/Repositories/PackageRepository.php
@@ -14,7 +14,6 @@ use BasePackage\Shared\Repositories\BaseRepository;
 use Modules\Subscription\Package\DTO\CreatePackageDTO;
 use Modules\Subscription\Package\DTO\UpdatePackageDTO;
 use Modules\Subscription\Package\Models\PackageFeature;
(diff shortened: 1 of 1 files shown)


modules/Subscription/Package/Repositories/PackageRepository.php
```diff
@@ -193,26 +192,13 @@ public function syncPermissions(Package $package, array $permissionIds, array $l
...
-            if ($limit === null) {
-                if ($permission->name === 'archive-library.archive-library*file.create') {
-                    $limit = 1000;
-                } elseif ($permission->name === 'archive-library.archive-library*folder.create') {
-                    $limit = 100;
-                }
-            }
-            
-            $syncData[$permission->id] = ['limit' => $limit];
+        foreach ($permissionIds as $permissionId) {
```
mabou7agar, line 198: The removal of the special handling for archive-library permissions seems like a breaking change. If these permissions still exist in the system, their default limits should be maintained. Consider either:
1. Keeping the special handling logic
2. Documenting this as a breaking change in the PR description
3. Moving the default limits to a configuration file

PR #1219, opened 2025-10-16
add limit by default

AmrSaleh2234:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/modules/Subscription/Package/Repositories/PackageRepository.php
+++ b/modules/Subscription/Package/Repositories/PackageRepository.php
@@ -6,7 +6,6 @@ namespace Modules\Subscription\Package\Repositories;
 
 use App\Exceptions\CustomException;
 use Illuminate\Support\Str;
-use Modules\RoleAndPermission\Models\Permission;
 use Ramsey\Uuid\UuidInterface;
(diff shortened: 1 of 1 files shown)


modules/Subscription/Package/Repositories/PackageRepository.php
```diff
@@ -192,26 +191,16 @@ public function syncPermissions(Package $package, array $permissionIds, array $l
...
-            if ($limit === null) {
-                if ($permission->name === 'archive-library.archive-library*file.create') {
-                    $limit = 1000;
-                } elseif ($permission->name === 'archive-library.archive-library*folder.create') {
-                    $limit = 100;
-                }
-            }
-
-            $syncData[$permission->id] = ['limit' => $limit];
+        foreach ($permissionIds as $permissionId) {
```
mabou7agar, line 198: The change removes the default limits for archive-library permissions (`archive-library.archive-library*file.create` and `archive-library.archive-library*folder.create`). This might be intentional, but we should confirm if these default limits should be completely removed or if they should be handled elsewhere in the code.

PR #1222, opened 2025-10-19
favourite

AmrSaleh2234:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change


modules/ArchiveLibrary/Folder/Controllers/FolderController.php
```diff
@@ -153,6 +154,7 @@ public function getFoldersAndFiles(GetFolderListRequest $request)
             $page,
             $perPage,
             $documentType,
+            $isFavourite,
```
mabou7agar, line 157: The parameter `$isFavourite` is added but not used in the method signature or documented. Ensure it's properly integrated into the method logic or remove it if unused.

modules/ArchiveLibrary/Folder/Repositories/FolderRepository.php
```diff
@@ -246,6 +248,13 @@ public function getFoldersAndFilesByParent(
             }
         }
 
+        // Filter by favourite status if provided
+        if ($isFavourite !== null && $isFavourite === true) {
```
mabou7agar, line 252: The condition `$isFavourite !== null && $isFavourite === true` can be simplified to just `$isFavourite === true` since we don't need to check for null if we're explicitly checking for true.

modules/ArchiveLibrary/Folder/Repositories/FolderRepository.php
```diff
@@ -246,6 +248,13 @@ public function getFoldersAndFilesByParent(
...
+            $filesQuery->whereHas('favouritedByUsers', function ($query) use ($userId) {
```
mabou7agar, line 253: Consider adding a check for `$userId` being null before using it in the query to prevent potential errors if `$userId` is not provided.

modules/ArchiveLibrary/Folder/Requests/GetFolderListRequest.php
```diff
@@ -27,7 +27,7 @@ public function rules(): array
 
     public function getDocumentType(): ?string
     {
-        return $this->input('document_type');
+        return $this->input('document_type') =="fav" ? null : $this->input('document_type');
```
mabou7agar, line 30: The comparison operator should use strict equality (`===`) instead of loose equality (`==`) to avoid type coercion issues.

modules/ArchiveLibrary/Folder/Requests/GetFolderListRequest.php
```diff
@@ -64,4 +64,9 @@ public function getSort(): ?string
     {
         return $this->input('sort');
     }
+
+    public function getIsFavourite(): ?bool
+    {
+        return $this->input('document_type') !== "fav" ? true : null;
```
mabou7agar, line 70: The logic here seems inverted. If the intention is to return `true` when the document type is "fav", the condition should be `return $this->input('document_type') === "fav" ? true : null;`. Also, consider using a strict comparison (`===`) for better type safety.

mabou7agar, line 70: The method name `getIsFavourite()` suggests it should return a boolean value, but it returns `null` in some cases. Consider making the return type consistent (always boolean) or clearly document when `null` can be returned.

modules/ArchiveLibrary/Folder/Services/FolderCRUDService.php
```diff
@@ -143,6 +144,7 @@ public function getFoldersAndFiles(
             $page,
             $perPage,
             $documentType,
+            $isFavourite,
```
mabou7agar, line 147: The parameter `$isFavourite` is added but there's no type hint or default value specified. Consider adding a type hint (e.g., `bool $isFavourite = false`) for better code clarity and type safety.

PR #1227, opened 2025-10-20
show without authintication

AmrSaleh2234:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/modules/ArchiveLibrary/File/Resources/routes/api.php
+++ b/modules/ArchiveLibrary/File/Resources/routes/api.php
@@ -3,6 +3,7 @@
 use Illuminate\Support\Facades\Route;
 use Modules\ArchiveLibrary\File\Controllers\FileController;
 use Modules\RoleAndPermission\Enums\Permission;
+use Stancl\Tenancy\Middleware\InitializeTenancyByRequestData;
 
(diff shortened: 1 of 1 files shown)


modules/ArchiveLibrary/File/Resources/routes/api.php
```diff
@@ -61,3 +57,4 @@
     Route::post('/{id}', [FileController::class, 'update'])
         ->permission(Permission::FILE_UPDATE());
 });
+Route::get('/{id}', [FileController::class, 'show'])->middleware(InitializeTenancyByRequestData::class);
```
mabou7agar, line 60: The new route for `show` should be consistent with the existing routes. Since other routes are wrapped in the `Route::group` with `middleware(['auth:sanctum'])`, this one should be as well to maintain consistency and security. Alternatively, if this is intentionally meant to be public, we should add a comment explaining why it's outside the auth group.

mabou7agar, line 60: The route is missing proper permission check like other routes (`Permission::FILE_VIEW()`). If this is meant to be a public route, we should explicitly state that in the code by either:
1. Adding `->withoutMiddleware([Authenticate::class])`
2. Or moving it outside the auth group with a clear comment

PR #1231, opened 2025-10-20
omit middlware from request

AmrSaleh2234:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/modules/ArchiveLibrary/File/Repositories/FileRepository.php
+++ b/modules/ArchiveLibrary/File/Repositories/FileRepository.php
@@ -34,9 +34,9 @@ class FileRepository extends BaseRepository
 
     public function getFile(UuidInterface $id): File
     {
-        return $this->findOneByOrFail([
+        return $this->model->withoutTenancy->where([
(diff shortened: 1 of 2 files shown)


modules/ArchiveLibrary/File/Repositories/FileRepository.php
```diff
@@ -34,9 +34,9 @@ public function getFileList(?int $page, ?int $perPage = 10): Collection
 
     public function getFile(UuidInterface $id): File
     {
-        return $this->findOneByOrFail([
+        return $this->model->withoutTenancy->where([
```
mabou7agar, line 37: The change from `findOneByOrFail()` to `withoutTenancy->where()->first()` might have unintended consequences:
1. The `findOneByOrFail()` method typically throws an exception if no record is found, while `first()` returns null
2. The tenancy scope removal could expose data that should be tenant-restricted
3. There's no error handling for the case when file is not found

PR #1243, opened 2025-10-21
-

AbdeleahmanEl-Desoky:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change


modules/Ecommerce/EcoCategory/Requests/Dashboard/CreateEcoCategoryDashboardRequest.php
```diff
@@ -72,7 +72,7 @@ public function createCreateEcoCategoryDTO(): CreateEcoCategoryDashboardDTO
             companyId: Uuid::fromString(tenant("id")),
             name: $validatedData['name'],
             parentId: isset($validatedData['parent_id']) ? Uuid::fromString($validatedData['parent_id']) : null,
-            priority: $validatedData['priority'] ?? 0
+            priority: (int)($validatedData['priority'] ?? 0)
```
mabou7agar, line 75: The type casting to `(int)` is unnecessary here since the null coalescing operator `??` already ensures the value is 0 (an integer) when `priority` is not set. This change adds no value and could be reverted.

PR #1248, opened 2025-10-21
update file

AmrSaleh2234:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/modules/ArchiveLibrary/File/Requests/UpdateFileRequest.php
+++ b/modules/ArchiveLibrary/File/Requests/UpdateFileRequest.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Modules\ArchiveLibrary\File\Requests;
 
 use Illuminate\Foundation\Http\FormRequest;
+use Illuminate\Validation\Rule;
 use Ramsey\Uuid\Uuid;
 use Modules\ArchiveLibrary\File\Commands\UpdateFileCommand;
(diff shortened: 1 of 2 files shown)


modules/ArchiveLibrary/File/Requests/UpdateFileRequest.php
```diff
@@ -15,7 +16,14 @@ public function rules(): array
     {
         return [
             'name' => 'required|string|max:255',
-            'reference_number' => 'required|unique:files,reference_number,' . $this->route('id'),
+            'reference_number' => [
+                'required',
+                Rule::unique('files', 'reference_number')
```
mabou7agar, line 21: The `Rule::unique()` should be imported at the top of the file. Add `use Illuminate\Validation\Rule;` if not already present.

modules/ArchiveLibrary/File/Requests/UpdateFileRequest.php
```diff
@@ -15,7 +16,14 @@ public function rules(): array
...
+                    ->ignore($this->route('id'))
+                    ->where(function ($query) {
+                        return $query->where('company_id', tenant('id'));
```
mabou7agar, line 24: The `tenant('id')` call might throw an error if tenant is not set. Consider adding a null check or providing a default value.

PR #1251, opened 2025-10-21
Refactor/link

AmrSaleh2234:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/modules/Company/CompanyCore/DTO/CompanyProfile/CreateCompanyOfficialDocumentDTO.php
+++ b/modules/Company/CompanyCore/DTO/CompanyProfile/CreateCompanyOfficialDocumentDTO.php
@@ -12,7 +12,7 @@ class CreateCompanyOfficialDocumentDTO
     public function __construct(
         private ManagementHierarchy $managementHierarchy,
         private ?string        $name,
-        private string        $description,
+        private ?string        $description,
         private string        $documentNumber,
         private string        $startDate,
         private string        $endDate,
--- a/modules/Company/CompanyCore/Database/Migrations/2025_05_26_074107_update_company_official_documents_table.php
+++ b/modules/Company/CompanyCore/Database/Migrations/2025_05_26_074107_update_company_official_documents_table.php
@@ -1,0 +1,20 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Support\Facades\Schema;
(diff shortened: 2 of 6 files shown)


modules/Company/CompanyCore/DTO/CompanyProfile/CreateCompanyOfficialDocumentDTO.php
```diff
@@ -12,7 +12,7 @@ class CreateCompanyOfficialDocumentDTO
     public function __construct(
         private ManagementHierarchy $managementHierarchy,
         private ?string        $name,
-        private string        $description,
+        private ?string        $description,
```
mabou7agar, line 15: The change from `string` to `?string` for the `description` property should be documented in the class docblock to reflect that this property can now be null. While you mentioned not to suggest adding comments, this is a type change that affects the class contract.

modules/Company/CompanyCore/Database/Migrations/2025_05_26_074107_update_company_official_documents_table.php
```diff
@@ -0,0 +1,20 @@
...
+    {
+        Schema::table('company_official_documents', function (Blueprint $table) {
+            $table->text('description')->nullable()->change();
+
+        });
+    }
+
+    public function down(): void
+    {
+    }
```
mabou7agar, line 19: The `down()` method should implement the reverse of the `up()` method to properly support rollbacks. Currently it's empty which could cause issues if this migration needs to be rolled back.

modules/Company/CompanyCore/Repositories/CompanyLegalDataRepository.php
```diff
@@ -153,9 +153,9 @@ public function updateCompanyLegalData(array $data = [])
             }
             DB::commit();
 
-            if ($lastLegalData) {
-                event(new CompanyLegalDataUpdated($lastLegalData->fresh()));
-            }
+//            if ($lastLegalData) {
```
mabou7agar, line 156: Instead of commenting out the event dispatch, it would be better to either:
1. Remove it completely if it's no longer needed
2. Keep it active if it's still required

Commented code should not be committed to the repository as it adds noise and confusion.

PR #1258, opened 2025-10-22
-

AbdeleahmanEl-Desoky:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/modules/Ecommerce/EcoProduct/Presenters/Dashboard/EcoProductDashboardPresenter.php
+++ b/modules/Ecommerce/EcoProduct/Presenters/Dashboard/EcoProductDashboardPresenter.php
@@ -19,6 +19,7 @@ class EcoProductDashboardPresenter extends AbstractPresenter
 
     protected function present(bool $isListing = false): array
     {
+        $firstMedia = $this->ecoProduct->getFirstMedia('eco_product_main_image');
     
         return [
             'id' => $this->ecoProduct->id,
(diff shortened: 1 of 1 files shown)


modules/Ecommerce/EcoProduct/Presenters/Dashboard/EcoProductDashboardPresenter.php
```diff
@@ -19,6 +19,7 @@ public function __construct(EcoProduct $ecoProduct)
 
     protected function present(bool $isListing = false): array
     {
+        $firstMedia = $this->ecoProduct->getFirstMedia('eco_product_main_image');
```
mabou7agar, line 22: The variable `$firstMedia` is defined but never used in the return array. Either remove it or include it in the returned data structure if it's needed.

modules/Ecommerce/EcoProduct/Presenters/Dashboard/EcoProductDashboardPresenter.php
```diff
@@ -27,6 +28,8 @@ protected function present(bool $isListing = false): array
             'stock' => $this->ecoProduct->stock,
             'sku' => $this->ecoProduct->sku,
             'is_visible' => (int)$this->ecoProduct->is_visible,
+            'type' => $this->ecoProduct->type,
+            'main_image' => $firstMedia ? (new MediaPresenter($firstMedia))->getData() : null,
```
mabou7agar, line 32: Consider adding a null check for `$firstMedia` before passing it to `MediaPresenter` to ensure type safety. Alternatively, handle the case where `$firstMedia` might not be an expected type.

PR #1263, opened 2025-10-22
calc size another way

AmrSaleh2234:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/modules/ArchiveLibrary/Folder/Presenters/FolderPresenter.php
+++ b/modules/ArchiveLibrary/Folder/Presenters/FolderPresenter.php
@@ -175,6 +175,6 @@ class FolderPresenter extends AbstractPresenter
 
 
 
-        return (int) $totalSize+$totalSizeDirect;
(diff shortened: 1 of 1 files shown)


modules/ArchiveLibrary/Folder/Presenters/FolderPresenter.php
```diff
@@ -175,6 +175,6 @@ private function getFolderSize(): int
 
 
 
-        return (int) $totalSize+$totalSizeDirect;
+        return  $totalSize+$totalSizeDirect;
```
mabou7agar, line 178: The type casting `(int)` was removed. If this was intentional, please confirm that:
- The variables `$totalSize` and `$totalSizeDirect` are guaranteed to be integers
- There's no risk of floating point arithmetic here

If not, the type casting should be kept to ensure integer return type.

PR #1267, opened 2025-10-22
calc file size

AmrSaleh2234:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/modules/ArchiveLibrary/File/Observers/FileObserver.php
+++ b/modules/ArchiveLibrary/File/Observers/FileObserver.php
@@ -269,7 +269,7 @@ class FileObserver
     private function getFileSizeInMB(File $file)
     {
         // Try to get from Spatie media first (direct uploads)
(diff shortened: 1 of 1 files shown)


modules/ArchiveLibrary/File/Observers/FileObserver.php
```diff
@@ -269,7 +269,7 @@ public function deleting(File $file): void
     private function getFileSizeInMB(File $file)
     {
         // Try to get from Spatie media first (direct uploads)
-        $media = $file->getFirstMedia();
+        $media = $file->getFirstMedia("upload");
```
mabou7agar, line 272: The hardcoded string `"upload"` should be extracted to a constant or configuration value to avoid magic strings and make it easier to maintain if the media collection name needs to change in the future.

PR #1268, opened 2025-10-22
-

AbdeleahmanEl-Desoky:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/modules/Ecommerce/EcoProduct/Presenters/Dashboard/EcoProductDashboardDetailsPresenter.php
+++ b/modules/Ecommerce/EcoProduct/Presenters/Dashboard/EcoProductDashboardDetailsPresenter.php
@@ -65,7 +65,7 @@ class EcoProductDashboardDetailsPresenter extends AbstractPresenter
             
             // Tax and shipping
             'vat_percentage' => $this->ecoProduct->vat_percentage ? (float) $this->ecoProduct->vat_percentage : null,
-            'price_includes_vat' => (bool) $this->ecoProduct->price_includes_vat,
+            'price_includes_vat' => (int) $this->ecoProduct->price_includes_vat,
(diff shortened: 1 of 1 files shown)


modules/Ecommerce/EcoProduct/Presenters/Dashboard/EcoProductDashboardDetailsPresenter.php
```diff
@@ -65,7 +65,7 @@ protected function present(bool $isListing = false): array
             
             // Tax and shipping
             'vat_percentage' => $this->ecoProduct->vat_percentage ? (float) $this->ecoProduct->vat_percentage : null,
-            'price_includes_vat' => (bool) $this->ecoProduct->price_includes_vat,
+            'price_includes_vat' => (int) $this->ecoProduct->price_includes_vat,
```
mabou7agar, line 68: Changing from `(bool)` to `(int)` cast for `price_includes_vat` seems inconsistent with the field's likely boolean nature. If this is intentional, please ensure the database and business logic also treat this as an integer rather than boolean.

modules/Ecommerce/EcoProduct/Presenters/Dashboard/EcoProductDashboardDetailsPresenter.php
```diff
@@ -65,7 +65,7 @@ protected function present(bool $isListing = false): array
...
             'shipping_amount' => $this->ecoProduct->shipping_amount ? (float) $this->ecoProduct->shipping_amount : null,
             'shipping_included_in_price' => (int) $this->ecoProduct->shipping_included_in_price,
```
mabou7agar, line 70: If `price_includes_vat` is being changed to integer type, consider whether `shipping_included_in_price` should also be consistently handled (either both as boolean or both as integer).

PR #1278, opened 2025-10-23
make companyId as string

AmrSaleh2234:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/modules/CompanyUser/Services/CompanyUserCRUDService.php
+++ b/modules/CompanyUser/Services/CompanyUserCRUDService.php
@@ -59,6 +59,7 @@ class CompanyUserCRUDService
     public function sendEmailAssignToCompanyToUser($user , $companyId)
     {
         $userInCompany = $this->userRepository->findOneBy(["global_company_user_id" => $user->global_id])->first();
+        $companyId = (string)$companyId;
         $company = $this->companyRepository->getCompany(Uuid::fromString($companyId));
         $data = [
             "name" => $userInCompany->name,


modules/CompanyUser/Services/CompanyUserCRUDService.php
```diff
@@ -59,6 +59,7 @@ public function create(CreateCompanyUserDTO $createCompanyUserDTO, CreateCompany
     public function sendEmailAssignToCompanyToUser($user , $companyId)
     {
         $userInCompany = $this->userRepository->findOneBy(["global_company_user_id" => $user->global_id])->first();
+        $companyId = (string)$companyId;
```
mabou7agar, line 62: The type casting to string should be done at the method parameter level (`public function sendEmailAssignToCompanyToUser($user, string $companyId)`) rather than inside the method to ensure type consistency throughout the codebase.

modules/CompanyUser/Services/CompanyUserCRUDService.php
```diff
@@ -59,6 +59,7 @@ public function create(CreateCompanyUserDTO $createCompanyUserDTO, CreateCompany
...
         $company = $this->companyRepository->getCompany(Uuid::fromString($companyId));
```
mabou7agar, line 63: Since `$companyId` is now being cast to string, we should verify that `Uuid::fromString()` properly handles cases where the input might already be a string representation of a UUID to avoid potential duplicate string conversions.

PR #1315, opened 2025-10-28
fix findByType

AbdeleahmanEl-Desoky:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/modules/Ecommerce/Banner/Services/SettingPageCRUDService.php
+++ b/modules/Ecommerce/Banner/Services/SettingPageCRUDService.php
@@ -52,7 +52,7 @@ class SettingPageCRUDService
 
     public function getByCompanyAndType(UuidInterface $companyId, string $type): ?SettingPage
     {
-        return $this->repository->findByCompanyAndType($companyId, $type);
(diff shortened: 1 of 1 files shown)


modules/Ecommerce/Banner/Services/SettingPageCRUDService.php
```diff
@@ -52,7 +52,7 @@ public function get(UuidInterface $id): SettingPage
 
     public function getByCompanyAndType(UuidInterface $companyId, string $type): ?SettingPage
     {
-        return $this->repository->findByCompanyAndType($companyId, $type);
+        return $this->repository->findByType($type);
```
mabou7agar, line 55: The method signature changed but the method name `getByCompanyAndType` still implies it should take both company and type parameters. Either:
1. Keep the original implementation
2. Rename the method to `getByType` and remove the unused `$companyId` parameter
3. If company filtering is still needed, modify the repository method to include it

PR #1316, opened 2025-10-28
fix findByType

AbdeleahmanEl-Desoky:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/modules/Ecommerce/Banner/Services/SettingPageCRUDService.php
+++ b/modules/Ecommerce/Banner/Services/SettingPageCRUDService.php
@@ -23,8 +23,7 @@ class SettingPageCRUDService
     public function upsert(UpsertSettingPageDTO $upsertSettingPageDTO): SettingPage
     {
         // Check if setting page exists for this company and type
-        $existingSettingPage = $this->repository->findByCompanyAndType(
-            $upsertSettingPageDTO->companyId,
+        $existingSettingPage = $this->repository->findByType(
             $upsertSettingPageDTO->type
         );
 


modules/Ecommerce/Banner/Services/SettingPageCRUDService.php
```diff
@@ -23,8 +23,7 @@ public function __construct(
     public function upsert(UpsertSettingPageDTO $upsertSettingPageDTO): SettingPage
     {
         // Check if setting page exists for this company and type
-        $existingSettingPage = $this->repository->findByCompanyAndType(
-            $upsertSettingPageDTO->companyId,
+        $existingSettingPage = $this->repository->findByType(
```
mabou7agar, line 26: The change removes the companyId parameter from findByType call, but we need to ensure this won't cause issues with data isolation between companies. If this is intentional, consider adding validation elsewhere to prevent cross-company data access.

modules/Ecommerce/Banner/Services/SettingPageCRUDService.php
```diff
@@ -23,8 +23,7 @@ public function __construct(
...
             $upsertSettingPageDTO->type
```
mabou7agar, line 27: The parameter change from findByCompanyAndType to findByType suggests a significant logic change. Make sure all calling code is updated accordingly and tests are modified to verify the new behavior.

PR #1329, opened 2025-11-01
fix SocialMedia

AbdeleahmanEl-Desoky:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/modules/Ecommerce/EcoBrand/Repositories/EcoBrandRepository.php
+++ b/modules/Ecommerce/EcoBrand/Repositories/EcoBrandRepository.php
@@ -38,6 +38,10 @@ class EcoBrandRepository extends BaseRepository
             }])
             ->orderBy('created_at', 'desc');
 
+        if (method_exists($this->model, 'scopeFilter')) {
+            $query->filter(request()->all());
+        }
+
         $total = $query->count();
         
         if ($page && $perPage) {


modules/Ecommerce/EcoBrand/Repositories/EcoBrandRepository.php
```diff
@@ -38,6 +38,10 @@ public function getBrandListWithCounts(?int $page, ?int $perPage = 10): array
             }])
             ->orderBy('created_at', 'desc');
 
+        if (method_exists($this->model, 'scopeFilter')) {
```
mabou7agar, line 41: The `filter` method call should be wrapped in a try-catch block to handle potential exceptions from invalid request parameters.

modules/Ecommerce/EcoBrand/Repositories/EcoBrandRepository.php
```diff
@@ -38,6 +38,10 @@ public function getBrandListWithCounts(?int $page, ?int $perPage = 10): array
...
+            $query->filter(request()->all());
```
mabou7agar, line 42: Consider validating `request()->all()` before passing it to the filter scope to ensure only expected parameters are used for filtering.

PR #1331, opened 2025-11-01
Ecommerce

AbdeleahmanEl-Desoky:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change


modules/Ecommerce/Warehous/Models/Warehous.php
```diff
@@ -63,7 +65,7 @@ public function country()
 
     public function city()
     {
-        return $this->belongsTo(City::class);
+        return $this->belongsTo(State::class);
```
mabou7agar, line 68: The relationship changed from `City` to `State` but the method name `city()` is now misleading. Consider renaming the method to `state()` to accurately reflect the relationship.

PR #1332, opened 2025-11-01
Ecommerce

AbdeleahmanEl-Desoky:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/modules/Ecommerce/Warehous/Models/Warehous.php
+++ b/modules/Ecommerce/Warehous/Models/Warehous.php
@@ -65,7 +65,7 @@ class Warehous extends Model
 
     public function city()
     {
-        return $this->belongsTo(State::class);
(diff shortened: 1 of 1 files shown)


modules/Ecommerce/Warehous/Models/Warehous.php
```diff
@@ -65,7 +65,7 @@ public function country()
 
     public function city()
     {
-        return $this->belongsTo(State::class);
+        return $this->belongsTo(State::class,'city_id');
```
mabou7agar, line 68: The foreign key parameter should be consistent with the relationship naming. Since the method is named `city()`, the foreign key should ideally be named `city_id` to match Laravel's conventions. However, if the database column is actually named differently, this should be clarified.

PR #1333, opened 2025-11-01
Ecommerce

AbdeleahmanEl-Desoky:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/modules/Ecommerce/Warehous/Presenters/WarehousPresenter.php
+++ b/modules/Ecommerce/Warehous/Presenters/WarehousPresenter.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Modules\Ecommerce\Warehous\Presenters;
 
+use Modules\Country\Presenters\CountryStateCityPresenter;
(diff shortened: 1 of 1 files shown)


modules/Ecommerce/Warehous/Presenters/WarehousPresenter.php
```diff
@@ -26,7 +27,10 @@ protected function present(bool $isListing = false): array
             'name' => $this->warehous->name,
             'is_default' => (int) $this->warehous->is_default,
             'country' => $this->warehous->country ?( new CountryPresenter($this->warehous->country))->getData() : null,
-            'city' => $this->warehous->city ?( new CityPresenter($this->warehous->city))->getData() : null,
+            'city' => $this->warehous->city ?[
```
mabou7agar, line 30: The change from using `CityPresenter` to directly accessing city properties breaks consistency with how `country` is handled above. Either both should use presenters or both should access properties directly for consistency.

PR #1344, opened 2025-11-05
.

AmrSaleh2234:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change


modules/Audit/Models/Audit.php
```diff
@@ -40,4 +42,9 @@ protected static function newFactory(): AuditFactory
     {
         return AuditFactory::new();
     }
+
+    public function getRelationshipToPrimaryModel(): string
+    {
+        return "user";
```
mabou7agar, line 48: Consider using a constant or config value for the relationship name `"user"` instead of hardcoding it to make it more maintainable and configurable.

PR #1347, opened 2025-11-05
.

AmrSaleh2234:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/modules/Audit/Models/Audit.php
+++ b/modules/Audit/Models/Audit.php
@@ -10,6 +10,7 @@ use Illuminate\Database\Eloquent\Factories\HasFactory;
 use Modules\Audit\Database\factories\AuditFactory;
 use BasePackage\Shared\Traits\BaseFilterable;
 use Modules\User\Models\User;
+use Stancl\Tenancy\Database\Concerns\BelongsToPrimaryModel;
 
 //use BasePackage\Shared\Traits\HasTranslations;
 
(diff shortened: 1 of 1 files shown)


modules/Audit/Models/Audit.php
```diff
@@ -39,11 +41,16 @@ class Audit extends \OwenIt\Auditing\Models\Audit
     public function user()
 
     {
-        return $this->belongsTo(User::class , "user_id","id")->withoutTenancy();
+        return $this->belongsTo(User::class , "user_id","id");
```
mabou7agar, line 44: The `withoutTenancy()` scope was removed. Was this intentional? If multi-tenancy is being used, this might cause issues with data isolation between tenants.

modules/Audit/Models/Audit.php
```diff
@@ -39,11 +41,16 @@ class Audit extends \OwenIt\Auditing\Models\Audit
...
     }
 
     protected static function newFactory(): AuditFactory
     {
         return AuditFactory::new();
     }
+
+    public function getRelationshipToPrimaryModel(): string
```
mabou7agar, line 52: The method `getRelationshipToPrimaryModel()` returns a hardcoded string "user". Consider making this more dynamic if there might be other relationships in the future.

PR #1353, opened 2025-11-05
.

AmrSaleh2234:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/modules/Attendance/Models/Attendance.php
+++ b/modules/Attendance/Models/Attendance.php
@@ -51,11 +51,12 @@ use Modules\UserInfo\UserProfessionalData\Models\UserProfessionalData;
  * @property-read Collection|AttendanceBreak[] $breaks
  * @property-read AttendanceConstraint|null $attendanceConstraint
  */
-class Attendance extends Model
+class Attendance extends Model implements Auditable
 {
     use UuidTrait;
     use BaseFilterable;
     // use SoftDeletes;
+    use \OwenIt\Auditing\Auditable;
     use CustomBelongsToTenant;
 
     protected $table = 'attendances';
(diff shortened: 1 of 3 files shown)


modules/Attendance/Models/Attendance.php
```diff
@@ -51,11 +51,12 @@
  * @property-read Collection|AttendanceBreak[] $breaks
  * @property-read AttendanceConstraint|null $attendanceConstraint
  */
-class Attendance extends Model
+class Attendance extends Model implements Auditable
```
mabou7agar, line 54: The `Auditable` interface should be imported at the top of the file with other imports rather than being used directly in the class declaration.

modules/Attendance/Models/Attendance.php
```diff
@@ -51,11 +51,12 @@
...
 {
     use UuidTrait;
     use BaseFilterable;
     // use SoftDeletes;
+    use \OwenIt\Auditing\Auditable;
```
mabou7agar, line 59: The `Auditable` trait should be imported at the top of the file with other imports rather than being used directly in the class.

modules/Attendance/Models/AttendanceConstraint.php
```diff
@@ -39,7 +39,7 @@
  * @property-read User $creator
  * @property-read User $updater
  */
-class AttendanceConstraint extends Model
+class AttendanceConstraint extends Model implements Auditable
```
mabou7agar, line 42: The `Auditable` interface should be imported at the top of the file. Add `use OwenIt\Auditing\Contracts\Auditable;` to the imports section.

PR #1357, opened 2025-11-11
omit try and catch

AmrSaleh2234:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/modules/Company/ManagementHierarchy/Repositories/ManagementHierarchyRepository.php
+++ b/modules/Company/ManagementHierarchy/Repositories/ManagementHierarchyRepository.php
@@ -139,6 +139,7 @@ class ManagementHierarchyRepository extends BaseRepository
     {
         try {
             DB::beginTransaction();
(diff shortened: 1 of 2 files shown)


modules/Company/ManagementHierarchy/Repositories/ManagementHierarchyRepository.php
```diff
@@ -139,6 +139,7 @@ public function createBranch(array $branchData, array $addressData,): Management
     {
         try {
             DB::beginTransaction();
+            tenancy()->initialize((string)$branchData["company_id"]);
```
mabou7agar, line 142: The `tenancy()->initialize()` call is added without any error handling. If this fails, the transaction will be left open. Consider either:
1. Wrapping this in a try-catch and rolling back the transaction on failure
2. Moving it before the transaction starts if it's a prerequisite for the transaction

PR #1359, opened 2025-11-12
Bugfix/create comany from not central comapny

AmrSaleh2234:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/app/Scopes/CustomTenantScope.php
+++ b/app/Scopes/CustomTenantScope.php
@@ -31,7 +31,7 @@ class CustomTenantScope implements Scope
 
         if($model instanceof Company) //if model is Company and not central , apply the tenant filtering on id because no have company_id column
         {
-            $builder->where($model->qualifyColumn("id"), $tenant->id);
+//            $builder->where($model->qualifyColumn("id"), $tenant->id);
             return;
         }
 
--- a/modules/Company/ManagementHierarchy/Repositories/ManagementHierarchyRepository.php
+++ b/modules/Company/ManagementHierarchy/Repositories/ManagementHierarchyRepository.php
@@ -139,6 +139,7 @@ class ManagementHierarchyRepository extends BaseRepository
     {
         try {
             DB::beginTransaction();
+            tenancy()->initialize((string)$branchData["company_id"]);
             $managementHierarchy = $this->create($branchData + ["id" => $this->nextId]);
 
             $this->nextId = $this->nextId + 1;
--- a/modules/User/Repositories/UserRepository.php
+++ b/modules/User/Repositories/UserRepository.php
@@ -209,6 +209,33 @@ class UserRepository extends BaseRepository
         ];
     }
 
+    public function getEmployeeInCurrentCompany($page = 1, $perPage = 10)
+    {
+        $type = CompanyUserRole::EMPLOYEE->value;
+
+        if (method_exists($this->model, 'scopeFilter')) {
+            $query = $this->model->filter(request()->all());
+        } else {
+            $query = $this->model;
+        }
+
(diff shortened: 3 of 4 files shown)


app/Scopes/CustomTenantScope.php
```diff
@@ -31,7 +31,7 @@ public function apply(Builder $builder, Model $model): void
 
         if($model instanceof Company) //if model is Company and not central , apply the tenant filtering on id because no have company_id column
         {
-            $builder->where($model->qualifyColumn("id"), $tenant->id);
+//            $builder->where($model->qualifyColumn("id"), $tenant->id);
```
mabou7agar, line 34: The commented-out code should be removed completely rather than just being commented out. If this filtering is no longer needed, we should clean up the code.

app/Scopes/CustomTenantScope.php
```diff
@@ -31,7 +31,7 @@ public function apply(Builder $builder, Model $model): void
...
             return;
```
mabou7agar, line 35: Since we're removing the tenant filtering for Company model, we should verify if this `return` statement is still needed or if it might cause unintended behavior by skipping other scope conditions.

modules/Company/ManagementHierarchy/Repositories/ManagementHierarchyRepository.php
```diff
@@ -139,6 +139,7 @@ public function createBranch(array $branchData, array $addressData,): Management
     {
         try {
             DB::beginTransaction();
+            tenancy()->initialize((string)$branchData["company_id"]);
```
mabou7agar, line 142: The `tenancy()->initialize()` call should be wrapped in a try-catch block to handle potential exceptions that might occur during tenant initialization. Failure to do so could leave the database transaction in an inconsistent state.

modules/User/Repositories/UserRepository.php
```diff
@@ -209,6 +209,33 @@ public function getEmployeeInCurrentCompanyWith($page = 1, $perPage = 10)
...
+    {
+        $type = CompanyUserRole::EMPLOYEE->value;
+
+        if (method_exists($this->model, 'scopeFilter')) {
+            $query = $this->model->filter(request()->all());
+        } else {
+            $query = $this->model;
+        }
+
+        $query = $query->when($type != null, function ($query) use ($type) {
```
mabou7agar, line 222: The `when` condition `$type != null` is redundant since `$type` is always set to `CompanyUserRole::EMPLOYEE->value` on line 214 and will never be null.

modules/User/Repositories/UserRepository.php
```diff
@@ -209,6 +209,33 @@ public function getEmployeeInCurrentCompanyWith($page = 1, $perPage = 10)
...
+            $query->whereHas("companyUserCompanies", function ($query) use ($type) {
+                $query->where("company_users_companies.role", $type);
```
mabou7agar, line 224: Consider using a constant or enum value directly instead of the string `"company_users_companies.role"` to avoid potential typos and make the code more maintainable.

modules/User/Repositories/UserRepository.php
```diff
@@ -209,6 +209,33 @@ public function getEmployeeInCurrentCompanyWith($page = 1, $perPage = 10)
...
+            });
+        })->where("company_id", tenant("id"))->whereNot("email","admin@constrix.com")
```
mabou7agar, line 226: The hardcoded email `"admin@constrix.com"` should be moved to a configuration file or constant to make it easier to maintain and change if needed.

PR #1362, opened 2025-11-12
mmm

AmrSaleh2234:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/modules/CompanyUser/Repositories/CompanyUserRepository.php
+++ b/modules/CompanyUser/Repositories/CompanyUserRepository.php
@@ -280,7 +280,7 @@ class CompanyUserRepository extends BaseRepository
 
     public function createCompanyUser(array $companyUserData, array $companyRole, array $branches = null, array $address = null, array $clientDetail = null, array $brokerDetail = null)
     {
-        try {
+//        try {
             $phone = $this->getPhoneNumberInfo($companyUserData['phone']);
 
             DB::beginTransaction();
@@ -391,10 +391,10 @@ class CompanyUserRepository extends BaseRepository
 //
             DB::commit();
             return $companyUser;
-        } catch (\Exception $exception) {
-            DB::rollBack();
-            throw new CustomException($exception->getMessage(), 400);
-        }
+//        } catch (\Exception $exception) {
+//            DB::rollBack();
+//            throw new CustomException($exception->getMessage(), 400);
+//        }
     }
 
     public function assignRoleCompanyUser(UuidInterface $id, array $companyUserRoleData, array $branches = null): void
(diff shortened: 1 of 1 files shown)


modules/CompanyUser/Repositories/CompanyUserRepository.php
```diff
@@ -391,10 +391,10 @@ public function createCompanyUser(array $companyUserData, array $companyRole, ar
 //
             DB::commit();
             return $companyUser;
-        } catch (\Exception $exception) {
-            DB::rollBack();
-            throw new CustomException($exception->getMessage(), 400);
-        }
+//        } catch (\Exception $exception) {
```
mabou7agar, line 394: Commented out code should be removed. If the exception handling is no longer needed, delete these lines completely rather than leaving them commented out.

modules/CompanyUser/Repositories/CompanyUserRepository.php
```diff
@@ -534,10 +534,10 @@ private function findOrCreateUserInCompany(CompanyUser $companyUser, $companyId,
     private function handleOwnerPermissions(User $user, $companyId): void
     {
         if ($user->is_owner) {
-            $branch = $this->managementHierarchyRepository->findOneBy([
+            $branch = $this->managementHierarchyRepository->model->where([
```
mabou7agar, line 537: The change from `findOneBy` to `model->where()->first()` might have performance implications since `findOneBy` could be optimized at the ORM level. Consider checking if this change aligns with the intended performance characteristics.

modules/CompanyUser/Repositories/CompanyUserRepository.php
```diff
@@ -534,10 +534,10 @@ private function findOrCreateUserInCompany(CompanyUser $companyUser, $companyId,
...
                 "company_id" => $companyId,
                 "parent_id" => null
-            ]);
+            ])->first();
```
mabou7agar, line 540: The `first()` method call should be on a new line for better readability and consistency with the codebase style.

modules/CompanyUser/Repositories/CompanyUserRepository.php
```diff
@@ -548,12 +548,12 @@ private function handleOwnerPermissions(User $user, $companyId): void
...
             $branch->update(["manager_id" => $user->id]);
 
-            $this->managementHierarchyRepository->findOneBy([
+            $this->managementHierarchyRepository->model->where([
                 "company_id" => $companyId,
                 "parent_id" => $branch->id,
                 "type" => "management",
                 "is_main" => 1
-            ])->update(["manager_id" => $user->id]);
+            ])->first()->update(["manager_id" => $user->id]);
```
mabou7agar, line 556: The change from `findOneBy` to `where()->first()` might introduce a potential issue if no record is found. The original code would silently fail if no record exists, while the new code would throw an exception when trying to call `update()` on null. Consider adding a null check before calling `update()`.

modules/CompanyUser/Repositories/CompanyUserRepository.php
```diff
@@ -635,23 +635,23 @@ private function createMultiBranchAssociations(User $user, CompanyUserCompany $c
      */
     private function getMainBranchData($companyId, ?array $branches = null): array
     {
-        $mainBranchId = $this->managementHierarchyRepository->findOneBy([
+        $mainBranchId = $this->managementHierarchyRepository->model->withoutTenancy()->where([
             "company_id" => $companyId,
             "parent_id" => null
-        ])->id;
+        ])  ->first()->id;
```
mabou7agar, line 641: There's an extra space before `->first()->id` that should be removed for consistent formatting.

PR #1369, opened 2025-11-13
Feature/add role for user

AmrSaleh2234:
### Types of changes
- [ ] New feature 
- [ ] Bugfix 
- [ ] Refactoring 
- [ ] Code style update 
- [ ] Breaking change

### DIFF
--- a/modules/ArchiveLibrary/File/Repositories/FileRepository.php
+++ b/modules/ArchiveLibrary/File/Repositories/FileRepository.php
@@ -207,6 +207,7 @@ class FileRepository extends BaseRepository
                         null,
                         true
                     );
+                }
 
                     request()->files->set('file', $uploadedFile);
 
--- a/modules/ArchiveLibrary/File/Repositories/FileRepository.php
+++ b/modules/ArchiveLibrary/File/Repositories/FileRepository.php
@@ -196,6 +196,19 @@ class FileRepository extends BaseRepository
 
             foreach ($fileIds as $fileId) {
                 $originalFile = $this->getFile(\Ramsey\Uuid\Uuid::fromString($fileId));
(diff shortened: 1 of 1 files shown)


modules/ArchiveLibrary/File/Repositories/FileRepository.php
```diff
@@ -196,6 +196,20 @@ public function copyFile(array $fileIds, ?UuidInterface $targetFolderId): array
...
+                    $uploadedFile = new UploadedFile(
+                        $filePath,
+                        $media->file_name,
+                        $media->mime_type,
+                        null,
+                        true
+                    );
+                }
+
+                    request()->files->set('file', $uploadedFile);
```
mabou7agar, line 212: The indentation here is incorrect - it should be aligned with the `if ($media)` block above.

mabou7agar, line 212: There's no null check for `$uploadedFile` before using it. If `$media` is null, this will throw an error.

modules/ArchiveLibrary/File/Repositories/FileRepository.php
```diff
@@ -196,6 +196,20 @@ public function copyFile(array $fileIds, ?UuidInterface $targetFolderId): array
...
+
+                if ($media) {
+                    $filePath = $media->getPath();
+                    $uploadedFile = new UploadedFile(
+                        $filePath,
+                        $media->file_name,
+                        $media->mime_type,
+                        null,
+                        true
+                    );
```
mabou7agar, line 209: The `true` parameter for `$test` in `UploadedFile` constructor should be documented or explained since it's not obvious what it does.

PR #1456, opened 2025-12-10
handel issue in get data when delete

AmrSaleh2234:
## 🎯 Summary

- just handle relations 
## 🧾 Related Ticket

no related jira
## 🔄 Type of Change

- [ ] Feature
- [ ] Bugfix
- [ ] Refactor
- [x] Hotfix
- [x] Chore / Maintenance

## ✅ Checklist

- [x] Performed **self-review** of the code
- [ ] Updated **validations** if business logic changed
- [x] Updated **API docs / Postman collection** if the endpoint changed
- [ ] All **tests are passing** (if applicable)
- [x] No **breaking changes** for existing clients

## 🧱 Migration / Database Changes

- [x] No database changes
- [ ] There is a new migration (describe below):

PR #1457, opened 2025-12-10
Revert "handel issue in get data when delete"

AmrSaleh2234:
Reverts New-Vision-2050/Contsrix_Core_Api#1456

### DIFF
--- a/modules/CompanyUser/Models/CompanyUser.php
+++ b/modules/CompanyUser/Models/CompanyUser.php
@@ -105,8 +105,7 @@ class CompanyUser extends Model implements HasMedia
     public function companies()
     {
         return $this->belongsToMany(Company::class, 'company_users_companies', 'global_company_user_id', 'company_id')
-            ->withPivot('role', 'status')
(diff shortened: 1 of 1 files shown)


modules/CompanyUser/Models/CompanyUser.php
```diff
@@ -105,8 +105,7 @@ class CompanyUser extends Model implements HasMedia
     public function companies()
     {
         return $this->belongsToMany(Company::class, 'company_users_companies', 'global_company_user_id', 'company_id')
-            ->withPivot('role', 'status')
-            ->wherePivotNull('deleted_at');
+            ->withPivot('role', 'status');
```
YasserAlawadly, line 108: The removal of `->wherePivotNull('deleted_at')` might cause soft-deleted records to appear in the results. If this is intentional, please confirm. Otherwise, it should be kept to maintain consistent behavior with soft deletion.

YasserAlawadly (approved):

PR #1458, opened 2025-12-10
Revert "Revert "handel issue in get data when delete""

AmrSaleh2234:
Reverts New-Vision-2050/Contsrix_Core_Api#1457

### DIFF
--- a/modules/CompanyUser/Models/CompanyUser.php
+++ b/modules/CompanyUser/Models/CompanyUser.php
@@ -105,7 +105,8 @@ class CompanyUser extends Model implements HasMedia
     public function companies()
     {
         return $this->belongsToMany(Company::class, 'company_users_companies', 'global_company_user_id', 'company_id')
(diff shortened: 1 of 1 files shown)


modules/CompanyUser/Models/CompanyUser.php
```diff
@@ -105,7 +105,8 @@ class CompanyUser extends Model implements HasMedia
     public function companies()
     {
         return $this->belongsToMany(Company::class, 'company_users_companies', 'global_company_user_id', 'company_id')
-            ->withPivot('role', 'status');
+            ->withPivot('role', 'status')
+            ->wherePivotNull('deleted_at');
```
YasserAlawadly, line 109: The `wherePivotNull('deleted_at')` condition might unintentionally filter out records where `deleted_at` is not null but still need to be included in certain business logic scenarios. Consider if this is the intended behavior for all use cases of this relationship.

PR #1869, opened 2026-04-28
fix database

AbdeleahmanEl-Desoky:
## 🎯 Summary

-
## 🧾 Related Ticket

- Jira: [ABC-123](https://jira.your-company.com/browse/ABC-123)

## 🔄 Type of Change

- [ ] Feature
- [ ] Bugfix
- [ ] Refactor
- [ ] Hotfix
- [ ] Chore / Maintenance

## 🧪 How to Test

1. Call endpoint: `GET /api/...`
2. Use test data: `...`
3. Expected result:
   - 
    -

## ✅ Checklist

- [ ] Performed **self-review** of the code
- [ ] Updated **validations** if business logic changed
- [ ] Updated **API docs / Postman collection** if the endpoint changed
- [ ] All **tests are passing** (if applicable)
- [ ] No **breaking changes** for existing clients

## 🧱 Migration / Database Changes

- [ ] No database changes
- [ ] There is a new migration (describe below):

PR #1872, opened 2026-04-28
Revert "fix database"

AmrSaleh2234:
Reverts New-Vision-2050/Contsrix_Core_Api#1869


modules/ClientRequest/Database/Migrations/2025_02_24_000004_create_client_requests_table.php
```diff
@@ -26,8 +26,8 @@ public function up(): void
 
             $table->foreign('company_id')->references('id')->on('companies')->onDelete('cascade');
             $table->foreign('term_setting_id')->references('id')->on('term_settings')->onDelete('set null');
-            // FKs to management_hierarchies: added after that table exists — see
-            // 2025_03_11_000001_add_client_requests_management_hierarchy_foreign_keys.php
+            $table->foreign('branch_id')->references('id')->on('management_hierarchies')->onDelete('set null');
```
AmrSaleh2234, line 29: The foreign key constraints for `branch_id` and `management_id` should not be added here as they depend on the `management_hierarchies` table which is created later. This was correctly handled in the reverted version by having a separate migration for these foreign keys, which is the proper approach to avoid dependency issues.

modules/Project/ProjectManagement/Database/Migrations/2025_02_24_000001_add_contract_fields_to_projects_table.php
```diff
@@ -33,8 +33,13 @@ public function up(): void
             $table->index(['project_owner_type', 'project_owner_id']);
         });
 
-        // FK to management_hierarchies is added in a later migration so it runs after
-        // that table exists (see 2025_03_11_000000_projects_branch_id_foreign_key.php).
+        // Add foreign key constraints after column creation
```
AmrSaleh2234, line 36: The comment about FK to management_hierarchies being added in a later migration should not be removed as it provides important context about the migration sequence.

modules/Project/ProjectManagement/Database/Migrations/2025_02_24_000001_add_contract_fields_to_projects_table.php
```diff
@@ -33,8 +33,13 @@ public function up(): void
...
+        Schema::table('projects', function (Blueprint $table) {
```
AmrSaleh2234, line 37: The new code adds a foreign key constraint immediately rather than in a later migration as previously planned. This could cause issues if the management_hierarchies table doesn't exist yet. Consider keeping the original approach of adding this in a separate migration.

